PR #1, opened 2018-03-25
update configure.py

zeno40:


Maratyszcza:
It removes support of Android, 32-bit x86, and 32-bit ARM. Why?

zeno40:
Sorry, I made a mistake! I wanted to update the cpuinfo version I forked from your repo with the latest changes you've made. I looked for some button to synchronize and pushed 'pull request' as I thought it will do the trick. I'm still not fully at ease with github.

zeno40:
If removed build.target.is_x86 and build.target.is_android because I was getting an error saying: Platform instance has no attribute "is_x86" when I launched python ./configure.py after calling confu setup.

zeno40:
Thanks for the info. After updating confu all problems where solved!

PR #15, opened 2018-12-14
Fixed possible misspelling in error message

norahsakal:
This might be a misspelling; a small fix, however, it would make it easier to find other references to the same error.


Maratyszcza:
Thank you!

PR #22, opened 2019-03-16
Install the config

xsacha:


Maratyszcza:
Could you explain the details? Also, which CMake version did you target (cpuinfo currently targets CMake 2.8, but we can update it to 3.5. Going above 3.5, however, would be hard as some cpuinfo users target 3.5)

xsacha:
Definitely not above 3.5. I think 3.0.

xsacha:
Just a note: If you set the Travis CI to use the version of CMake you desire, then you never need to worry about compatibility -- it would fail to compile.
It is compiling so I suppose Travis CI CMake is new enough.

PR #29, opened 2020-01-21
Upstream the cpuinfo updates in XNNPACK/third_party/cpuinfo.patch

AshkanAliabadi:
Author: The XNNPACK Team.

### DIFF
--- a/include/cpuinfo.h
+++ b/include/cpuinfo.h
@@ -38,8 +38,16 @@
 	#define CPUINFO_ARCH_PNACL 1
 #endif
 
-#if defined(EMSCRIPTEN)
+#if defined(__asmjs__)
 	#define CPUINFO_ARCH_ASMJS 1
+#endif
+
+#if defined(__wasm__)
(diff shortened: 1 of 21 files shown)


include/cpuinfo.h
```diff
@@ -1682,6 +1747,11 @@ uint32_t CPUINFO_ABI cpuinfo_get_l2_caches_count(void);
 uint32_t CPUINFO_ABI cpuinfo_get_l3_caches_count(void);
 uint32_t CPUINFO_ABI cpuinfo_get_l4_caches_count(void);
 
+/**
+ * Returns upper bound on cache size.
+ */
+uint32_t CPUINFO_ABI cpuinfo_get_max_cache_size(void);
```
AshkanAliabadi: @Maratyszcza This function is not defined as far as I can tell.

Maratyszcza: Looks like I missed it when creating a patch. The implementation is trivial: just return ‘cpuinfo_max_cache_size’

Issue #36, opened 2020-03-30
pytorch build fails: __NR_getcpu undeclared, a possible fix included


pearu:
The current master branch of pytorch fails to build with the following error:
```
[86/3487] Building C object confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/api.c.o
FAILED: confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/api.c.o 
/home/pearu/miniconda3/envs/pytorch-cuda-dev/bin/x86_64-conda_cos6-linux-gnu-cc -DCPUINFO_LOG_LEVEL=2 -DTH_BLAS_MKL -D_GNU_SOURCE=1 -I../third_party/cpuinfo/src -I../third_party/cpuinfo/include -I../third_party/cpuinfo/deps/clog/include -isystem ../third_party/protobuf/src -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/pearu/miniconda3/envs/pytorch-cuda-dev/include -I/usr/local/cuda-10.1.243/include -L/home/pearu/miniconda3/envs/pytorch-cuda-dev/lib -O3 -DNDEBUG -fPIC   -std=c99 -MD -MT confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/api.c.o -MF confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/api.c.o.d -o confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/api.c.o   -c ../third_party/cpuinfo/src/api.c
In file included from ../third_party/cpuinfo/src/cpuinfo/internal-api.h:11:0,
                 from ../third_party/cpuinfo/src/api.c:5:
../third_party/cpuinfo/src/api.c: In function 'cpuinfo_get_current_processor':
../third_party/cpuinfo/src/api.c:316:31: error: '__NR_getcpu' undeclared (first use in this function); did you mean '__NR_getcwd'?
   if CPUINFO_UNLIKELY(syscall(__NR_getcpu, &cpu, NULL, NULL) != 0) {
                               ^
../third_party/cpuinfo/src/cpuinfo/common.h:16:59: note: in definition of macro 'CPUINFO_UNLIKELY'
  #define CPUINFO_UNLIKELY(condition) (__builtin_expect(!!(condition), 0))
                                                           ^~~~~~~~~
[...]

PR #37, opened 2020-04-05
Fix a build error some users are experiencing wherein __NR_getcpu is not defined

AshkanAliabadi:
https://github.com/pytorch/cpuinfo/issues/36

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -10,6 +10,9 @@
 
 	#include <unistd.h>
 	#include <sys/syscall.h>
+	#if !defined(__NR_getcpu)
+		#include <asm-generic/unistd.h>
+	#endif
 #endif
 
 bool cpuinfo_is_initialized = false;


src/api.c
```diff
@@ -10,6 +10,7 @@
 
 	#include <unistd.h>
 	#include <sys/syscall.h>
+	#include <asm-generic/unistd.h>
```
pearu, line 13: To minimize possible conflicts for systems where inclusion of `unistd.h` already defines `__NR_getcpu`, I suggest:
```
	#if !defined(__NR_getcpu)
		#include <asm-generic/unistd.h>
	#endif
```
With this change, I confirm that pytorch master builds successfully here (Ubuntu 18.04, using conda environment).

rgommers:
In addition to what @pearu said, I would change line 314 from `#ifdef __linux__` to:
```
if defined(__linux__) && defined(__NR_getcpu)
```
That will fix things on older kernels that don't have `__NR_getcpu` at all as well.

Maratyszcza:
I second @pearu's suggestion to include `asm-generic/unistd.h` only if the `unistd.h` doesn't have the right symbol. However, the suggestion in @rgommers's comment is very dangerous: it introduce the risk of silently regressing performance (because `cpuinfo_get_current_*` functions return default result) without any indication of failure. `getcpu` syscall was introduced in Linux 2.6.19 released on Nov 29, 2006 so we can just assume that it is always supported.

src/api.c
AshkanAliabadi: Thanks.  Regarding your question on what PyTorch tests to run - it is difficult to say.  cpuinfo is used both directly inside PyTorch and by its dependencies.  Many codepaths in PyTorch are conditionally enabled or disabled with CMake variables and preprocessor directives in the code, and there are also different platforms to consider so running the CI in its entirety is a must in my opinion.

Maratyszcza:
LGTM

rgommers:
> However, the suggestion in @rgommers's comment is very dangerous: it introduce the risk of silently regressing performance

I don't think that's true, since it can only fail if the symbol is missing completely, i.e. for old kernels. Also, pretty much every use of `__NR_getcpu` in public code (including in Tensorflow [here](https://github.com/tensorflow/tensorflow/blob/fa3b55e940cef475c8691487f4c00ac900a67d04/tensorflow/core/platform/default/port.cc#L100)) is protected like that. Within 2 weeks of merging the unprotected version, multiple people ran into build issues. 

That said, there was no response yet from @bryan-lunt on gh-36, and this does fix things for @pearu and me, so merging this PR as-is is an improvement. It's still likely to come back as runtime errors once people on older clusters install the next release though.

AshkanAliabadi:
@driess Can you please stamp this?

rgommers:
A ping on this, can someone merge this PR please?

rgommers:
Thanks! PR to update submodule in PyTorch at https://github.com/pytorch/pytorch/pull/37083

PR #38, opened 2020-04-14
Set unknow uarch default l3 cache info

Danliran:
Serval project depends on cpuinfo project to detect the CPU cache size, but if a cpu arch is undefined
will cause some procedure function exception, it`s better to set default l3 cache info.

### DIFF
--- a/src/arm/cache.c
+++ b/src/arm/cache.c
@@ -1470,6 +1470,11 @@ void cpuinfo_arm_decode_cache(
 					.associativity = 8,
 					.line_size = 64
 				};
+				*l3 = (struct cpuinfo_cache) {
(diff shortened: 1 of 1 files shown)


soumith:
@Maratyszcza can you review

Danliran:
@Maratyszcza many vendors designed CPU for server based on ARM core, such as Huawei Kunpeng 920 serials CPU based on ARMv8.2a,  this serials CPU have 32M L3 cache. please check  #39  . so I proposal NNPACK should detect cache size, not hard code cache size in init_hwinfo function for ARM. https://github.com/Maratyszcza/NNPACK/issues/183

PR #41, opened 2020-05-17
Android NDK 21

kamgurgul:
To build this library with NDK 21 two changes needs to be done:
* bump minimum platform version to 16
* drop support for armeabi 

Apart of that there was one missing file in `Android.mk`: `cache.c`.

If you don't want to support new NDK then this PR can safely be closed :)


Maratyszcza:
LGTM

PR #42, opened 2020-05-22
Mingw crosscompile fix

syoyo:
This simple fix solves compiling cpuinfo with MinGW gcc/clang cross-compiling on Linux(with case-sensitive filesystem). 

According to this,

https://stackoverflow.com/questions/15466613/lowercase-windows-h-and-uppercase-windows-h-difference

Using lowercase Win32  header filename should be always safe and would be recommended.


Maratyszcza:
LGTM

syoyo:
:100:

PR #46, opened 2020-07-30
fix(uclibc): fix uclibc build error

daimagou:
cpuinfo build will fail when use uclibc, try to solve it like this.
```
fatal error: sys/auxv.h: No such file or directory
#include <sys/auxv.h>
```


daimagou:
Please kindly review, in addition, is there any other better solution?

PR #48, opened 2020-11-18
Fix build for Apple Silicon

malfet:
MacOS machines based on Apple M1 silicon are identified by cmake as "arm64"

Modify build rules accordingly to recognize "arm64" is valid CPU configuration for cpuinfo

Found while working on https://github.com/pytorch/pytorch/issues/48145


src/arm/mach/init.c
```diff
@@ -101,6 +105,9 @@ static enum cpuinfo_uarch decode_uarch(uint32_t cpu_family, uint32_t cpu_subtype
 		case CPUFAMILY_ARM_LIGHTNING_THUNDER:
 			/* Hexa-core: 2x Lightning + 4x Thunder; Octa-core (presumed): 4x Lightning + 4x Thunder */
 			return core_index + 4 < core_count ? cpuinfo_uarch_lightning : cpuinfo_uarch_thunder;
+		case CPUFAMILY_ARM_FIRESTORM_ICESTORM:
+			/* Hexa-core: 2x Firestorm + 4x Icestorm; Octa-core (presumed): 4x Firestorm + 4x Icestorm */
```
Maratyszcza, line 109: Nit: this is not presumed, we know this is the case on M1

malfet, line 109:
```suggestion
			/* Hexa-core: 2x Firestorm + 4x Icestorm; Octa-core: 4x Firestorm + 4x Icestorm */
```

PR #58, opened 2021-06-25
Fix fp16 arithmetic detection for Cortex-X1 device

jcao-ai:
FP16 arithmetic detection for Cortext-X1 is not covered. 

CPUs of type Cortext-X1: Snapdragon 888, Samsung Exynos 2100


Maratyszcza:
The AArch32 part is superseded by #168. AArch64 part is presumably unnecessary, as all devices with Cortex-X1 have a recent enough Linux kernel to report FP16 support via `getauxval(AT_HWCAP)`.

PR #59, opened 2021-07-03
add loongarch support

nanxiongchao:


include/cpuinfo.h
```diff
@@ -255,6 +259,9 @@ enum cpuinfo_vendor {
 	 * Sold its ARM designs in 1997. The last processor design was released in 1997.
 	 */
 	cpuinfo_vendor_dec       = 57,
+
+	/** Loongson Technologies, Inc. Vendor of LOONGARCH processor microarchitectures. */
+	cpuinfo_vendor_loongarch     = 258,
```
xen0n, line 264: You're putting the new value in the "defunct CPU vendors" section, are you shutting your production yet? If not, that's a mistake, and you should be inserting the definition in the section above, not here.

Also this list is representing companies, so this should be "loongson" instead of "loongarch", BTW the "LOONGARCH" should really be "LoongArch".

tools/cpu-info.c
```diff
@@ -42,6 +42,8 @@ static const char* vendor_to_string(enum cpuinfo_vendor vendor) {
 			return "Broadcom";
 		case cpuinfo_vendor_apm:
 			return "Applied Micro";
+		case cpuinfo_vendor_loongarch:
+			return "LoongArch";
```
xen0n, line 46: This is concerning vendors, not architectures, so this should be "Loongson" too.

include/cpuinfo.h
```diff
@@ -46,6 +46,10 @@
 	#endif
 #endif
 
+#if defined(__loongarch__) && defined(__loongarch64)
```
xen0n, line 49: Only matching against `__loongarch64` is enough, there's no configuration that has `__loongarch64` but is not LoongArch.

And the section below needs update too, you need a `#ifndef / #define CPUINFO_ARCH_LOONGARCH64 0` pair like all other arches.

xen0n (changes requested):
Is these changes enough? For example, I see definition for `CPUINFO_ARCH_LOONGARCH64` but never see any usage of that. I haven't looked very closely at the project as a whole, but maybe you should post some minimal test results to let maintainers confirm the changes are actually correct?

PR #62, opened 2021-07-15
Remove HOMEPAGE_URL from cmake project definition

tkonolige:
HOMEPAGE_URL in cmake's project function is only supported by cmake >= 3.12. In order to support versions down to 3.5, it needs to be removed.


Maratyszcza:
@AshkanAliabadi PTAL

PR #75, opened 2022-01-04
cpuinfo: aarch64: add cache configuration details for neoverse-n1/v1/n2

snadampal:


src/arm/cache.c
```diff
@@ -1238,6 +1238,59 @@ void cpuinfo_arm_decode_cache(
...
+                         *
+                         * A7.1 About the L2 memory system
+                         *   The L2 memory subsystem consist of:
+			 *    - An 8-way set associative L2 cache with a configurable size of 256KB, 512KB, or 1024KB. Cache lines
+			 *      have a fixed length of 64 bytes.
+                         *    - Strictly inclusive with L1 data cache.
+			 *    - When configured with instruction cache hardware coherency, strictly inclusive with L1 instruction cache.
+			 *    - When configured without instruction cache hardware coherency, weakly inclusive with L1 instruction cache.
+                         */
+                        const uint32_t l2_size = 1 * 1024 * 1024;
```
Maratyszcza, line 1269: This should be the lower bound for cache size, i.e. 256K

src/arm/cache.c
```diff
@@ -1238,6 +1238,59 @@ void cpuinfo_arm_decode_cache(
...
+                        const uint32_t l3_size = 32 * 1024 * 1024;
```
Maratyszcza, line 1270: This should be a conservative estimate too. Underestimating cache size is much better for performance of cache blocking algorithms than overstimating.

Maratyszcza (changes requested):

src/arm/cache.c
snadampal: I think the ideal way is to read the hw config and report the accurate cache sizes, will look into it. I wonder why the implementation specific cache sizes are hardcoded in this library.

snadampal:
Hi @Maratyszcza, I have updated the PR to read sizes from kernel populated cache indices for L2 and L3 cached. I'm falling back to the hardcoded defaults if kernel doesn't provide them.

snadampal:
Hi @Maratyszcza, please let me know if you have any comments on these new commits. Thank you!

snadampal:
Please let me know if there are any comments. I'm happy to address them.

src/arm/cache.c
```diff
@@ -1,11 +1,15 @@
...
 
 #include <cpuinfo.h>
 #include <cpuinfo/internal-api.h>
 #include <cpuinfo/log.h>
 #include <arm/api.h>
 #include <arm/midr.h>
 
+#define STRINGIFY(token) #token
+#define CACHESIZE_FILENAME_FORMAT "/sys/devices/system/cpu/cpu%" PRIu32 "/cache/index%" PRIu32 "/size"
+#define CACHESIZE_FILENAME_SIZE (sizeof("/sys/devices/system/cpu/cpu" STRINGIFY(UINT32_MAX) "/cache/index%" STRINGIFY(UINT32_MAX) "/size"))
```
malfet, line 12: This feels like a wrong way to put those changes, as "/sys" filesystem make sense only for Linux OS

src/arm/cache.c
```diff
@@ -1238,6 +1243,84 @@ void cpuinfo_arm_decode_cache(
...
+								CACHESIZE_FILENAME_FORMAT, system_processor_id, 2);
+
+			const int chars_formatted_l3 = snprintf(l3_cachesize_filename, CACHESIZE_FILENAME_SIZE,
+								CACHESIZE_FILENAME_FORMAT, system_processor_id, 3);
+
+			p = fopen(l2_cachesize_filename, "r");
+			if (p) {
+				fscanf(p, "%d", &l2_size_KB);
+				fclose(p);
+			}
```
malfet, line 1293: Please use `cpuinfo_linux_parse_small_file`

malfet (changes requested):
Please use `cpuinfo_linux_parse_small_file` and move OS specific changes to `arm/linux/...` folder

snadampal:
@malfet , Updated the PR with your feedback, please review. Thank you!

src/arm/cache.c
```diff
@@ -1239,6 +1241,66 @@ void cpuinfo_arm_decode_cache(
...
+                         *    - Strictly inclusive with L1 data cache.
+			 *    - When configured with instruction cache hardware coherency, strictly inclusive with L1 instruction cache.
+			 *    - When configured without instruction cache hardware coherency, weakly inclusive with L1 instruction cache.
+                         */
+
+			const uint32_t min_l2_size_KB= 256;
+			const uint32_t min_l3_size_KB = 0;
+
+			uint32_t proc_l2_size_KB = cpuinfo_linux_get_processor_cache_size(system_processor_id, 2);
+			uint32_t proc_l3_size_KB = cpuinfo_linux_get_processor_cache_size(system_processor_id, 3);
```
malfet, line 1279: This code would break compilation for say iOS or M1 platforms, wouldn't it?

malfet (changes requested):
Your changes would break compilation on non-Linux arm platforms, would they not (as `src/arm/cache.c` should not have any OS specific code. 
Do you think it would be better if you could split this PR into 2:
 - One that adds new CPU types and checks for bfloat16 and sve_bfloat16 (which fits well with the current model of the library)
 - Another that adds the logic of dynamically queuing cache sizes for neonverse family of SoC, where @Maratyszcza can perhaps provide some guidance how it best could be done

src/linux/processors.c
```diff
@@ -215,6 +218,26 @@ bool cpuinfo_linux_get_processor_package_id(uint32_t processor, uint32_t package
...
+                cpuinfo_log_warning("failed to format filename for cache size of processor %"PRIu32, processor);
+                return 0;
+        }
+
+        uint32_t cache_size;
+        if (cpuinfo_linux_parse_small_file(cache_size_filename, CACHE_FILESIZE, uint32_parser, &cache_size)) {
+                cpuinfo_log_debug("parsed cache size value of %"PRIu32" KB for level %"PRIu32" for logical processor %"PRIu32" from %s",cache_size, cache_level, processor, cache_size_filename);
+                return cache_size;
+        } else {
+                cpuinfo_log_info("failed to parse cache size for level %"PRIu32"  for processor %"PRIu32" from %s",
```
malfet, line 235: Shouldn't this be an error or at least a warning?
```suggestion
                cpuinfo_log_error("failed to parse cache size for level %"PRIu32"  for processor %"PRIu32" from %s",
```

snadampal: Since I'm anyway falling back to the conservative configuration if the device doesn't publish it, I left it as just info log. And I don't think this is an error, because kernel doesn't mandate it, and there are devices with cache info not populated in acpi or dts tables in kernel.

src/arm/cache.c
snadampal: Sure, I will split it into 2 PRs.
Regarding the dynamic cache querying feature, having a general framework support is good as it is going to benefit all SoCs, not just the neoverse family.

malfet (approved):

PR #81, opened 2022-03-06
Add support for Bazel build system

Vertexwahn:
This PR helps users of the Bazel build system to integrate cpuinfo in their builds.

Supporting an additional build system is of course more effort to maintain but helps to spread cpuinfo.
In the worst case the Bazel build does simply not work - in the best case, someone who uses Bazel can benefit from it.

Also, TensorFlow has a dependency on cpuinfo. TesorFlow build could be simplified if this PR is accepted.


malfet (approved):

malfet (approved):

.bazelversion
```diff
@@ -0,0 +1 @@
+5.0.0
```
Maratyszcza, line 1: Why is this particular version of Bazel pinned?

Vertexwahn: @Maratyszcza  It was the version of Bazel I used to test this change. In the meantime, I tested all versions of Bazel up to 6.3.2 successfully.

We could express this using skylib:

```
load("@bazel_skylib//lib:versions.bzl", "versions")

versions.check("5.0.0", "6.3.2")
```
see e.g. [here].(https://github.com/Vertexwahn/BazelDemos/blob/main/core_principles/bazel_version/WORKSPACE.bazel)

or just get rid of the `.bazelversion` file.

My main intention was just to document which version of Bazel was used to test this - we could also add a comment to the `.bazelversion` file:

```
5.0.0
# This was the version which was used to test those changes. Should also work with newer versions of Bazel.
```

This works, since Bazelisk only reads the first line (there is no real support for comments in `.bazelversion` files)

Could be that this change is broken with Bzlmod which is maybe introduced in Bazel 7.0.0.

In the end there is no real need for exact Bazel 5.0.0 and probably this works also with Bazel 2.0.0 and earlier...

Vertexwahn: @Maratyszcza - so there are 4 options:
- keep it as it is
- add a comment to `.bazelversion` file why this specific version and what other version work...
- remove the .bazelversion file
- use skylib for version check (but this needs an update every time a new Bazel version is released)

PR #85, opened 2022-03-27
Remove deps/ from git ignore file

Vertexwahn:
The `deps` folder contains clog which is a mandatory third-party lib of cpuinfo and should therefore not be ignored. Also updated Bazel version.


Vertexwahn:
@malfet Can you have a look?

Vertexwahn:
@malfet Can you approve?

Maratyszcza:
The `deps` is intentionally in `.gitignore` to avoid committing `clog` build objects and artifacts when the latter is configured/built within `cpuinfo` tree. Whenever `clog` is modified (extremely rare), it is added to commit via `git add -f ...`.

PR #89, opened 2022-05-26
Define namespace prefixed alias for cpuinfo in the CMake build

petrhosek:
This allows projects that depend on cpuinfo to use namespace qualified
target name regardless of whether they consume cpuinfo through
add_subdirectory or find_package.


Maratyszcza (approved):

petrhosek:
Thank you for the review, would it be possible to merge this PR?

Maratyszcza:
I don't have write access to the repo. @malfet PTAL

malfet:
@Maratyszcza does the change look good to you?
@petrhosek can you please share link to CMake doc explaining what namespace prefixes are and what version of cmake they got introduced

petrhosek:
It's a common convention in modern CMake to provide aliases for each installed library to ensure that the library can be consumes via `target_link(<pkg>::<target_name>)` both if the project is used via `find_package(<pkg>)` or via `add_subdirectory`.

See for example https://github.com/abseil/abseil-cpp/blob/d2c5297a3c3948de765100cb7e5cccca1210d23c/CMake/AbseilHelpers.cmake#L329. This can avoid similar constructs in downstream projects such as https://github.com/google/ruy/blob/368db7123bf3178c92c8502696c6865fb07bfcda/CMakeLists.txt#L85.

There's an ongoing discussion about how to supports this directly in CMake, see https://gitlab.kitware.com/cmake/cmake/-/issues/22687

malfet (approved):

malfet:
Thank you for the explanation, looks good to me

Maratyszcza:
cc @bjacob

PR #101, opened 2022-07-12
Use macOS 12 / iOS 15 perflevel sysctls for better cache info on Apple Platforms

kulinseth:
…e platforms


src/arm/mach/init.c
```diff
@@ -128,7 +182,7 @@ static enum cpuinfo_uarch decode_uarch(uint32_t cpu_family, uint32_t cpu_subtype
 	#endif
 }
 
-static void decode_package_name(char* package_name) {
+static void decode_hw_machine_package_name(char* package_name) {
```
Maratyszcza, line 185: This change looks unrelated. Please revert or move into a separate PR.

src/arm/mach/init.c
```diff
@@ -14,6 +14,12 @@
 #include <cpuinfo/internal-api.h>
 #include <cpuinfo/log.h>
 
+#define SAFE_FREE_NULL(x) 	\
```
Maratyszcza, line 17: This macro needs to go. cpuinfo use a the following pattern to ensure allocations properly deallocated:

```c
void some_function(args) {
  struct local_structure* some_local = NULL;

  some_local = allocate_local_structure(...);

  if (...something that may fail has failed...) {
    // log error
    goto cleanup;
  }
 
  // Commit changes to global state and return without invoking cleanup code
  global_structure /* e.g. cpuinfo_processors */ = local_structure; 
  return; // early return - skip cleanup

cleanup:
  free(local_structure);
}
```

src/arm/mach/init.c
```diff
@@ -54,6 +60,54 @@ struct cpuinfo_arm_isa cpuinfo_isa = {
 #endif
 };
 
+struct cache_array {
```
Maratyszcza, line 63: This looks redundant, just pass `struct cpuinfo_cache*` and `uint32_t count` as two separate variables.

src/arm/mach/init.c
```diff
@@ -54,6 +60,54 @@ struct cpuinfo_arm_isa cpuinfo_isa = {
...
+	struct cpuinfo_cache *caches;
+	uint32_t count;
+};
+
+/*
+ * iOS 15 and macOS Monterey 12 added sysctls to describe configuration information
+ * where not all cores are the same (number of cores, cache sizes).
+ * 
+ * Each perflevel sysctl has a prefix of `hw.perflevel??.` where ?? is the
```
Maratyszcza, line 72: Are perflevels reported by OS ordered? cpuinfo orders cores and logical processors from the most performant to the least performance, so `cpuinfo_get_core(0)` is the the fastest core and `cpuinfo_get_core(cpuinfo_get_cores_count() - 1)` is the most power-efficient core.

Maratyszcza (changes requested):
Please rebase on the current top-of-tree. Current diff is hard to read.

Maratyszcza (changes requested):
Changes to `cache-info` look unrelated, they should go into a separate PR.

PR #125, opened 2023-01-13
Add build support for PPC

Flamefire:
Add a `config_setting` for the "ppc"-cpu and include the common sources for linux, similar to the other "exotic" architectures.


Maratyszcza:
cpuinfo doesn't support PPC, and adding build config doesn't address it.

PR #129, opened 2023-01-17
Refactor logging and remove dependency on clog

Maratyszcza:


wangkuiyi:
Thank you for removing the dep on clog. Should we remove the source dir https://github.com/pytorch/cpuinfo/tree/main/deps/clog as well? @Maratyszcza

PR #133, opened 2023-01-29
Enable CXX only when needed for tests

jameshilliard:
We shouldn't require CXX support in the project as it's only needed for tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.5 FATAL_ERROR)
 # ---[ Setup project
 PROJECT(
     cpuinfo
-    LANGUAGES C CXX
+    LANGUAGES C
     )
 
 # ---[ Options.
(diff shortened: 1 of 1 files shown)


Maratyszcza (changes requested):
Benchmarks also require C++

jameshilliard:
> Benchmarks also require C++

Enabled C++ for benchmarks as well.

Maratyszcza (approved):

Maratyszcza:
CI build fails, PTAL

jameshilliard:
> CI build fails, PTAL

Yeah, needed to move the `ENABLE_LANGUAGE` block down a bit after `CPUINFO_SUPPORTED_PLATFORM` is defined/set.

CMakeLists.txt
```diff
@@ -103,6 +103,12 @@ ELSEIF(NOT CMAKE_SYSTEM_NAME MATCHES "^(Windows|WindowsStore|CYGWIN|MSYS|Darwin|
   ENDIF()
 ENDIF()
 
+IF(CPUINFO_SUPPORTED_PLATFORM)
+  IF(CPUINFO_BUILD_UNIT_TESTS OR CPUINFO_BUILD_BENCHMARKS)
```
Maratyszcza, line 107: Check `CPUINFO_BUILD_MOCK_TESTS` as well

jameshilliard: added

PR #135, opened 2023-02-12
Remove workaround for Wine

mrpippy:
Wine has supported `GetMaximumProcessorCount()` since version 6.10 (June 2021).

### DIFF
--- a/src/x86/windows/init.c
+++ b/src/x86/windows/init.c
@@ -95,15 +95,6 @@ static void cpuinfo_x86_count_caches(
 	*l4_count_ptr  = l4_count;
 }
 
-static bool cpuinfo_x86_windows_is_wine(void) {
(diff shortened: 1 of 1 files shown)


src/x86/windows/init.c
```diff
@@ -132,7 +122,7 @@ BOOL CALLBACK cpuinfo_x86_windows_init(PINIT_ONCE init_once, PVOID parameter, PV
 		x86_processor.topology.core_bits_offset + x86_processor.topology.core_bits_length);
 
 	/* WINE doesn't implement GetMaximumProcessorGroupCount and aborts when calling it */
```
Maratyszcza, line 124: Remove the comment as well

Maratyszcza (changes requested):

Maratyszcza (approved):

PR #137, opened 2023-02-21
Added Allwinner 2013 naming scheme support for A-Series.

kolyneh:
Allwinner updated their naming scehme since 2013:
[Allwinner_SoC_Family](https://linux-sunxi.org/Allwinner_SoC_Family)

The new scheme is `sunxiwxpx`,  the old scheme `sunxi` is preserved in this PR for compatibility.


malfet (changes requested):
Why are build system scripts changes in this PR?
Also, can you please split it into two:
 - One that adds new property
 - One that adds new chipset (could be this PR, based on the name)

kolyneh:
Thank you, the build system scripts changes in my local machine i'll fix it.

PR #141, opened 2023-03-23
Improve Windows Unicode support

richard-townsend-arm:
This PR adds support for the Snapdragon Compute Platform used in the Windows Dev Kit 2023 (aka Project Volterra). It also fixes an issue that showed up when enumerating the cache structures and improves string handling to work even if the UTF-16 UNICODE header is set. Please take a look.

### DIFF
--- a/src/arm/windows/init-by-logical-sys-info.c
+++ b/src/arm/windows/init-by-logical-sys-info.c
@@ -833,8 +833,18 @@ static bool connect_packages_cores_clusters_by_processors(
 
 		if (chip_info) {
 			size_t converted_chars = 0;
-			wcstombs_s(&converted_chars, package->name, CPUINFO_PACKAGE_NAME_MAX,
-			            chip_info->chip_name_string, _TRUNCATE);
+			if (!WideCharToMultiByte(
+					CP_UTF8,
+					WC_ERR_INVALID_CHARS,
+					chip_info->chip_name_string,
+					-1,
+					package->name,
+					CPUINFO_PACKAGE_NAME_MAX,
+					NULL,
+					NULL)) {
+				cpuinfo_log_error("cpu name character conversion error");
+				return false;
+			};
 		}
 
 		/* Set start indexes and counts per packages / clusters / cores - going backwards */
--- a/src/arm/windows/init-by-logical-sys-info.c
+++ b/src/arm/windows/init-by-logical-sys-info.c
@@ -832,8 +832,9 @@ static bool connect_packages_cores_clusters_by_processors(
 		processor->cluster = cluster;
 
 		if (chip_info) {
-			strncpy_s(package->name, CPUINFO_PACKAGE_NAME_MAX, chip_info->chip_name_string,
-				strnlen(chip_info->chip_name_string, CPUINFO_PACKAGE_NAME_MAX));
+			size_t converted_chars = 0;
+			wcstombs_s(&converted_chars, package->name, CPUINFO_PACKAGE_NAME_MAX,
+			            chip_info->chip_name_string, _TRUNCATE);
 		}
 
 		/* Set start indexes and counts per packages / clusters / cores - going backwards */
--- a/src/arm/windows/init.c
+++ b/src/arm/windows/init.c
@@ -21,24 +21,24 @@ static bool get_system_info_from_registry(
 	enum cpuinfo_vendor* vendor);
 
 struct vendor_info {
-	char vendor_name[VENDOR_NAME_MAX];
+	wchar_t vendor_name[VENDOR_NAME_MAX];
 	enum cpuinfo_vendor vendor;
 };
 
 /* Please add new vendor here! */
 static struct vendor_info vendors[] = {
 	{
-		"Qualcomm",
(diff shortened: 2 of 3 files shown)


src/arm/windows/init.c
```diff
@@ -101,8 +116,8 @@ bool get_core_uarch_for_efficiency(
 /* Static helper functions */
 
 static bool read_registry(
-	LPCTSTR subkey,
-	LPCTSTR value,
+	LPCSTR subkey,
```
Maratyszcza, line 119: Please use wide char types, as they are native to Windows

Maratyszcza: Yes, use `LPCWSTR`, log with `%ls`

Blackhex:
Please take a look also on this PR #144. We can somewhat merge them together...

richard-townsend-arm:
@Blackhex: let's proceed with merging your PR (#144) and refocus this one on Unicode support. 

One major omission from this PR is the ability to handle unknown / future CPUs (it'd be good to fall back to an "Unknown CPU" value if we can), as well as Hyper-V. It's proving slow to get the revisions to this PR approved inside Arm.

Blackhex:
@richard-townsend-arm Is this what you had in mind 20b4c43? It still fails the tests for unknown registry value but at least they don't crash on access violation or similar...

richard-townsend-arm:
Hi folks, apologies it's taken me a little while to follow up on this one - this PR is now exclusively about wide character support. I see the CLA bot is still unhappy (will chase up on that), in the meantime PTAL.

src/arm/windows/init.c
```diff
@@ -172,13 +172,13 @@ static bool read_registry(
 	if (*textBuffer) {
 		HeapFree(heap, 0, *textBuffer);
 	}
-	*textBuffer = HeapAlloc(heap, HEAP_ZERO_MEMORY, dataSize);
+	*textBuffer = HeapAlloc(heap, HEAP_ZERO_MEMORY, dataSize << 1);
```
Maratyszcza, line 175: Don't use shift, multiply by `sizeof(wchar_t)` instead

src/arm/windows/init.c
```diff
@@ -146,8 +146,8 @@ bool get_core_uarch_for_efficiency(
 /* Static helper functions */
 
 static bool read_registry(
-	LPCTSTR subkey,
-	LPCTSTR value,
+	LPCWSTR subkey,
+	LPCWSTR value,
 	char** textBuffer)
```
Maratyszcza, line 151: Should be text_buffer

src/arm/windows/init.c
```diff
@@ -146,8 +146,8 @@ bool get_core_uarch_for_efficiency(
 /* Static helper functions */
 
 static bool read_registry(
-	LPCTSTR subkey,
-	LPCTSTR value,
+	LPCWSTR subkey,
+	LPCWSTR value,
 	char** textBuffer)
 {
 	DWORD keyType = 0;
```
Maratyszcza, line 153: Should be key_type

src/arm/windows/init.c
richard-townsend-arm: Addressed in next revision

src/arm/windows/init.c
richard-townsend-arm: Addressed in next revision

src/arm/windows/init.c
richard-townsend-arm: Addressed in next revision.

src/arm/windows/init-by-logical-sys-info.c
```diff
@@ -832,8 +832,9 @@ static bool connect_packages_cores_clusters_by_processors(
 		processor->cluster = cluster;
 
 		if (chip_info) {
-			strncpy_s(package->name, CPUINFO_PACKAGE_NAME_MAX, chip_info->chip_name_string,
-				strnlen(chip_info->chip_name_string, CPUINFO_PACKAGE_NAME_MAX));
+			size_t converted_chars = 0;
+			wcstombs_s(&converted_chars, package->name, CPUINFO_PACKAGE_NAME_MAX,
```
Maratyszcza, line 836: I prefer WinAPI calls over CRT, as cpuinfo may be compiled with non-Microsoft compilers (e.g. Cygwin). Please use `WideCharToMultiByte` instead of `wcstombs_s`.

Maratyszcza (changes requested):
Thank you for the changes! One last fix, and it is ready to go!

src/arm/windows/init-by-logical-sys-info.c
richard-townsend-arm: Hi, changed to use `WideCharToMultiByte`: I think I've done it correctly and it seems to pass the tests (in Unicode and non-Unicode variants), but I think you're familiar with these things than I am. Could you take a look at `2b261b9`?

richard-townsend-arm:
Hi @Maratyszcza: any feedback on this latest iteration?

malfet (approved):
Looks good to me...

PR #147, opened 2023-05-11
Enable tvOS and watchOS

wangkuiyi:
I realized that we need this when I was building the OpenXLA/IREE runtime for tvOS and watchOS.

1. Before this change, the `CMakeLists.txt` file uses CMake built-in variable `IOS`, which is equivalent to `CMAKE_SYSTEM_NAME STREQUAL "iOS"`, to tell if it is building for Apple mobile platforms. This trick prevents CMake from recognizing tvOS and watchOS. 

   This pull requests changes the use of `IOS` into `set(APPLE_SYSTEM CMAKE_SYSTEM_NAME MATCHES "^(Darwin|iOS|tvOS|watchOS)")`.

2. Before this change, the `CMakeLists.txt` assumes that Apple's ARM64 platform has the name "arm64". However, M2 chip has the name "arm64e". This pull request changes `MATCHES "arm64"` into `MATCHES "arm64.*"`.


CMakeLists.txt
```diff
@@ -182,7 +184,10 @@ IF(CPUINFO_SUPPORTED_PLATFORM)
       ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64)$")
         LIST(APPEND CPUINFO_SRCS src/arm/linux/aarch64-isa.c)
       ENDIF()
-    ELSEIF(IOS OR (CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CPUINFO_TARGET_PROCESSOR STREQUAL "arm64"))
+    ELSEIF(IOS OR
```
wconstab, line 187: would it be possible to rename `IOS` to `APPLE_MOBILE_OS` and then set `APPLE_OS` = (iOS or tvOS or watchOS), and avoid having to check the specific strings in multiple places?

wangkuiyi: Regular expression matching came into my mind. Does it make sense?

wconstab: i was mostly trying to find a way to put the (ios or watchos or tvos) logic in one place and store that into a variable and then everywhere else just check the variable.

That's unless, of course, you wanted different behavior for the different os's somewhere...

wangkuiyi: Thank you for the clarification @wconstab . Done. Please take another look. Thanks!

wconstab: looks better to me! thanks.

Let me know if CI has issues that aren't your fault- gh actions has been unstable recently.  If its passing CI I think its good to land.  cc @malfet

wconstab:
@malfet anything else we need to check besides CI? (unfamiliar with cpuinfo lib)

CMakeLists.txt
```diff
@@ -55,9 +55,12 @@ MACRO(CPUINFO_TARGET_RUNTIME_LIBRARY target)
   ENDIF()
 ENDMACRO()
 
+SET(APPLE_SYSTEM CMAKE_SYSTEM_NAME MATCHES "^(Darwin|iOS|tvOS|watchOS)$")
+message(${APPLE_SYSTEM})
+
 # -- [ Determine target processor
 SET(CPUINFO_TARGET_PROCESSOR "${CMAKE_SYSTEM_PROCESSOR}")
-IF(IOS OR (CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_OSX_ARCHITECTURES MATCHES "^(x86_64|arm64)$"))
+IF(APPLE_SYSTEM)
```
malfet, line 63: One should not execute this path if `CMAKE_OSX_ARCHITECTURES` is empty.
```suggestion
IF(IS_APPLE_OS AND CMAKE_OSX_ARCHITECTURES MATCHES "^(x86_64|arm64.*)$")
```

CMakeLists.txt
```diff
@@ -82,20 +85,21 @@ IF(NOT CMAKE_SYSTEM_PROCESSOR)
...
 
+
 IF(NOT CMAKE_SYSTEM_NAME)
     MESSAGE(WARNING
       "Target operating system is not specified. "
       "cpuinfo will compile, but cpuinfo_initialize() will always fail.")
   SET(CPUINFO_SUPPORTED_PLATFORM FALSE)
 ELSEIF(NOT CMAKE_SYSTEM_NAME MATCHES "^(Windows|WindowsStore|CYGWIN|MSYS|Darwin|Linux|Android)$")
-  IF(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.14" AND NOT CMAKE_SYSTEM_NAME STREQUAL "iOS")
+  IF(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.14" AND NOT APPLE_SYSTEM)
```
malfet, line 102:
```suggestion
  IF(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.14" AND NOT IS_APPLE_OS)
```

CMakeLists.txt
```diff
@@ -153,7 +157,7 @@ IF(CPUINFO_SUPPORTED_PLATFORM)
       LIST(APPEND CPUINFO_SRCS
         src/x86/linux/init.c
         src/x86/linux/cpuinfo.c)
-    ELSEIF(CMAKE_SYSTEM_NAME STREQUAL "Darwin" OR CMAKE_SYSTEM_NAME STREQUAL "iOS")
+    ELSEIF(APPLE_SYSTEM)
```
malfet, line 160:
```suggestion
    ELSEIF(IS_APPLE_OS)
```

CMakeLists.txt
```diff
@@ -182,7 +186,7 @@ IF(CPUINFO_SUPPORTED_PLATFORM)
       ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64)$")
         LIST(APPEND CPUINFO_SRCS src/arm/linux/aarch64-isa.c)
       ENDIF()
-    ELSEIF(IOS OR (CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CPUINFO_TARGET_PROCESSOR STREQUAL "arm64"))
+    ELSEIF(APPLE_SYSTEM AND CPUINFO_TARGET_PROCESSOR MATCHES "arm64.*")
```
malfet, line 189:
```suggestion
    ELSEIF(IS_APPLE_OS AND CPUINFO_TARGET_PROCESSOR MATCHES "arm64.*")
```

CMakeLists.txt
```diff
@@ -202,7 +206,7 @@ IF(CPUINFO_SUPPORTED_PLATFORM)
       src/linux/multiline.c
       src/linux/cpulist.c
       src/linux/processors.c)
-  ELSEIF(CMAKE_SYSTEM_NAME STREQUAL "Darwin" OR CMAKE_SYSTEM_NAME STREQUAL "iOS")
+  ELSEIF(APPLE_SYSTEM)
```
malfet, line 209:
```suggestion
  ELSEIF(IS_APPLE_OS)
```

malfet (commented):
What's your test plan? 
Changed `APPLE_SYSTEM` to `IS_APPLE_OS` and removed debug print

CMakeLists.txt
```diff
@@ -55,9 +55,13 @@ MACRO(CPUINFO_TARGET_RUNTIME_LIBRARY target)
   ENDIF()
 ENDMACRO()
 
+-- [ Determine whether building for Apple's desktop or mobile OSes
```
malfet, line 58:
```suggestion
# -- [ Determine whether building for Apple's desktop or mobile OSes
```

CMakeLists.txt
```diff
@@ -55,9 +55,13 @@ MACRO(CPUINFO_TARGET_RUNTIME_LIBRARY target)
   ENDIF()
 ENDMACRO()
 
+# -- [ Determine whether building for Apple's desktop or mobile OSes
+SET(IS_APPLE_OS CMAKE_SYSTEM_NAME MATCHES "^(Darwin|iOS|tvOS|watchOS)$")
+
```
malfet, line 60:
```suggestion
```

wangkuiyi:
@wconstab @malfet - I verified that the change works on macOS M2 cross-compiling for macOS, iOS, iOS simulator, tvOS, and tvOS simulator with x86_64, arm64, and arm64e (M2).

I do not have macOS-x86_64, Linux, Windows, or Android. So I will have to rely on CI for these platforms.

malfet:
Ok, Windows and linter failures does not happen on trunk, so looks like this change is doing something unexpected, looking...

CMakeLists.txt
```diff
@@ -55,9 +55,12 @@ MACRO(CPUINFO_TARGET_RUNTIME_LIBRARY target)
   ENDIF()
 ENDMACRO()
 
+# -- [ Determine whether building for Apple's desktop or mobile OSes
+SET(IS_APPLE_OS CMAKE_SYSTEM_NAME MATCHES "^(Darwin|iOS|tvOS|watchOS)$")
```
malfet, line 59:
```suggestion
IF(CMAKE_SYSTEM_NAME MATCHES "^(Darwin|iOS|tvOS|watchOS)$")
  SET(IS_APPLE_OS TRUE)
ELSE()
  SET(IS_APPLE_OS FALSE)
ENDIF()
```

malfet (approved):

CMakeLists.txt
wangkuiyi: Thank you @malfet

PR #148, opened 2023-05-12
Added support for RISC-V architecture on Linux

GlassOfWhiskey:


GlassOfWhiskey:
Hi,
Any comment on this?

appujee:
ping. any update on when this will land?

include/cpuinfo.h
```diff
@@ -34,6 +34,16 @@
 	#define CPUINFO_ARCH_PPC64 1
 #endif
 
+#if defined(__riscv)
+    #if __riscv_xlen == 32
+        #define CPUINFO_ARCH_RISCV 1
```
topperc, line 39: Should this define be `CPUINFO_ARCH_RISCV32`? Unlike X86 and ARM, RISCV32 and RISCV64 were developed at the same time so 64 isn't an addition.

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,311 @@
...
+        } else if(*feature == 'd') {
+            #if CPUINFO_ARCH_RISCV || CPUINFO_ARCH_RISCV64
+                processor->features |= CPUINFO_RISCV_LINUX_FEATURE_D;
+            #endif
+        } else if(*feature == 'f') {
+            #if CPUINFO_ARCH_RISCV || CPUINFO_ARCH_RISCV64
+                processor->features |= CPUINFO_RISCV_LINUX_FEATURE_F;
+            #endif
+        } else if(*feature == 'i') {
+            #if CPUINFO_ARCH_RISCV
```
topperc, line 76: Why does this line not have `CPUINFO_ARCH_RISCV64`?

src/riscv/linux/hwcap.c
```diff
@@ -0,0 +1,37 @@
...
+    #define AT_HWCAP2 26
+#endif
+
+#if CPUINFO_MOCK
+    static uint32_t mock_hwcap = 0;
+    void cpuinfo_set_hwcap(uint32_t hwcap) {
+        mock_hwcap = hwcap;
+    }
+
+	static uint32_t mock_hwcap2 = 0;
```
topperc, line 20: Looks like this is indented with a tab character and the above is indented with spaces.

src/riscv/linux/riscv-isa.c
```diff
@@ -0,0 +1,29 @@
...
+    if (features & CPUINFO_RISCV_LINUX_FEATURE_C) {
+        isa->c = true;
+    }
+    if (features & CPUINFO_RISCV_LINUX_FEATURE_D) {
+        isa->d = true;
+    }
+    if (features & CPUINFO_RISCV_LINUX_FEATURE_F) {
+        isa->f = true;
+    }
+    #if CPUINFO_ARCH_RISCV
```
topperc, line 21: Why is this feature qualified with `CPUINFO_ARCH_RISCV`?

tools/cpu-info.c
```diff
@@ -42,6 +42,8 @@ static const char* vendor_to_string(enum cpuinfo_vendor vendor) {
 			return "Broadcom";
 		case cpuinfo_vendor_apm:
 			return "Applied Micro";
+        case cpuinfo_vendor_sifive:
```
topperc, line 45: Looks like this indented with spaces, but the surrounding lines use tabs?

tools/cpu-info.c
```diff
@@ -273,6 +275,8 @@ static const char* uarch_to_string(enum cpuinfo_uarch uarch) {
 			return "Dhyana";
 		case cpuinfo_uarch_taishan_v110:
 			return "TaiShan v110";
+        case cpuinfo_uarch_u74_mc:
```
topperc, line 278: Looks like this indented with spaces, but the surrounding lines use tabs?

appujee:
https://github.com/torvalds/linux/commit/16252e018a30486eedcfec81fc313445cac25bea maybe helpful in revising this patch.

Maratyszcza (changes requested):
Reformat consistently with the rest of cpuinfo. Use tabs for indentation.

GlassOfWhiskey:
@malfet I've seen that a recent PR (#190) on RISC-V support has been merged into the main branch. However, I think this PR can still be useful for some reasons:

 - It adds support for some additional features, like caches and uarchs;
 - It allows to retrieve CPU information by parsing the `/proc/cpuinfo` output, as a baseline method;
 - It disables the `hwprobe` path for kernels older than `6.4.0`, which prevents the actual version of `cpuinfo` to compile on almost all current RISC-V systems on the market.
 
It is still not clear to me why @prashanthswami chose to open a sibling PR instead of converging the efforts on a single one. However, I tried to manually merge #190 into this PR, applying some minor modifications when deemed necessary.
Please let me know what you think.

malfet:
@GlassOfWhiskey can you please amend the title and expand description to nightlight what this PR is trying to accomplish.

tools/isa-info.c
```diff
@@ -178,15 +178,20 @@ int main(int argc, char** argv) {
...
 #endif
+
 #if CPUINFO_ARCH_RISCV32 || CPUINFO_ARCH_RISCV64
 	printf("Instruction sets:\n");
-		printf("\tBase Integer: %s\n", cpuinfo_has_riscv_i() ? "yes" : "no");
-		printf("\tInteger Multiply/Divide: %s\n", cpuinfo_has_riscv_m() ? "yes" : "no");
-		printf("\tAtomics: %s\n", cpuinfo_has_riscv_a() ? "yes" : "no");
+		#if CPUINFO_ARCH_RISCV32
+			printf("\tBase Integer Instruction Set - 32-bit: %s\n", cpuinfo_has_riscv_i() ? "yes" : "no");
+		#elif CPUINFO_ARCH_RISCV64
```
malfet, line 186: Please remove `#elif there,  unless you are willing to add `#else #error` path

tools/isa-info.c
```diff
@@ -178,15 +178,20 @@ int main(int argc, char** argv) {
 		printf("\tPMULL: %s\n", cpuinfo_has_arm_pmull() ? "yes" : "no");
 		printf("\tCRC32: %s\n", cpuinfo_has_arm_crc32() ? "yes" : "no");
 #endif
+
 #if CPUINFO_ARCH_RISCV32 || CPUINFO_ARCH_RISCV64
 	printf("Instruction sets:\n");
-		printf("\tBase Integer: %s\n", cpuinfo_has_riscv_i() ? "yes" : "no");
-		printf("\tInteger Multiply/Divide: %s\n", cpuinfo_has_riscv_m() ? "yes" : "no");
-		printf("\tAtomics: %s\n", cpuinfo_has_riscv_a() ? "yes" : "no");
+		#if CPUINFO_ARCH_RISCV32
```
malfet, line 184: Nested preprocessor statements should not be indented
```suggestion
#if CPUINFO_ARCH_RISCV32
```

CMakeLists.txt
```diff
@@ -210,15 +210,18 @@ IF(CPUINFO_SUPPORTED_PLATFORM)
...
       LIST(APPEND CPUINFO_SRCS
+        src/riscv/linux/cpuinfo.c
+        src/riscv/linux/hwcap.c
         src/riscv/linux/init.c
-	src/riscv/linux/riscv-hw.c
-	src/riscv/linux/riscv-isa.c)
-    ENDIF()
+        src/riscv/linux/riscv-hw.c
+        src/riscv/linux/riscv-isa.c)
+    ENDIF ()
```
malfet, line 224: Why are you adding white-ace here?
```suggestion
    ENDIF()
```

CMakeLists.txt
```diff
@@ -210,15 +210,18 @@ IF(CPUINFO_SUPPORTED_PLATFORM)
       LIST(APPEND CPUINFO_SRCS
         src/arm/android/properties.c)
     ENDIF()
-  ELSEIF(CPUINFO_TARGET_PROCESSOR MATCHES "^(riscv(32|64))$")
+  ELSEIF (CPUINFO_TARGET_PROCESSOR MATCHES "^(riscv(32|64)?)$")
```
malfet, line 213: Why whitespace? And please explain why change of regex is necessary
```suggestion
  ELSEIF (CPUINFO_TARGET_PROCESSOR MATCHES "^(riscv(32|64)?)$")
```

src/linux/processors.c
```diff
@@ -88,7 +88,7 @@ inline static bool is_whitespace(char c) {
 	static const uint32_t default_max_processors_count = CPU_SETSIZE;
 #endif
 
-static bool uint32_parser(const char* text_start, const char* text_end, void* context) {
+static bool uint32_parser(const char* filename, const char* text_start, const char* text_end, void* context) {
```
malfet, line 91: This makes sense, but do you mind submitting a separate PR to accomplish that?

malfet (changes requested):
- Please stick to the same formatting rules
- Refactor `cpuinfo_smallfile_callback` into the separate PR

prashanthswami:
@GlassOfWhiskey - I sincerely apologize, this was poor GitHub practice on my part. I had written my PR a few months ago on an internal fork and needed it up streamed somewhat quickly, so I pushed it without analyzing the open PRs. When I looked at this issue last, it hadn't been active for a few months, so I was worried it would be inappropriate to ping here and ask for an accelerated landing, just to satisfy my own project timelines. In hindsight, I see that it would have been courteous to message here and coordinate, or CC you on the other PR. I'd like to make sure I don't do this again - should I CC you on any open PRs I have for RISC-V contributions to this repo?

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+    for (const char* digit_ptr = processor_start; digit_ptr != processor_end; digit_ptr++) {
+        const uint32_t digit = (uint32_t) (*digit_ptr - '0');
+        if (digit > 10) {
+            cpuinfo_log_warning("non-decimal suffix %.*s in /proc/cpuinfo processor number is ignored",
+                                (int) (processor_end - digit_ptr), digit_ptr);
+            break;
+        }
+
+        processor_number = processor_number * 10 + digit;
+    }
```
prashanthswami, line 37: Do we consider non-decimal suffixes not suspect? E.g. This would let something like "processor: +++100" through, which should give us pause, right? We're probably reading garbage and shouldn't assume it means 100. To that end, if we don't do any particular corrective action here, couldn't we replace this function with `strtoul` ?

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+    return processor_number;
+}
+
+static void parse_isa(
+        const char* isa_start,
+        const char* isa_end,
+        struct cpuinfo_riscv_linux_processor processor[restrict static 1])
+{
+    const size_t isa_length = (size_t) (isa_end - isa_start);
+    if (!(memcmp(isa_start, "rv32", 4) == 0 || memcmp(isa_start, "rv64", 4) == 0)) {
```
prashanthswami, line 48: Should we fail ahead of the memcmp if `isa_length < 4`?

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+            #endif
+        } else if(*feature == 'm') {
+            #if CPUINFO_ARCH_RISCV32 || CPUINFO_ARCH_RISCV64
+                processor->features |= CPUINFO_RISCV_LINUX_FEATURE_M;
+			#endif
+		} else if(*feature == 'v') {
+			#if CPUINFO_ARCH_RISCV32 || CPUINFO_ARCH_RISCV64
+				processor->features |= CPUINFO_RISCV_LINUX_FEATURE_V;
+			#endif
+		} else {
```
prashanthswami, line 87: Tab format.

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+                          (int) (isa_length), isa_start);
+        return;
+    }
+
+    /* Mark the features as valid */
+    processor->flags |= CPUINFO_RISCV_LINUX_VALID_FEATURES | CPUINFO_RISCV_LINUX_VALID_PROCESSOR;
+
+    const char* feature = isa_start + 4;
+    for (;feature != isa_end; feature++) {
+        if(*feature == 'a') {
```
prashanthswami, line 59: Same nitpick as above, can we order these semantically to match the RISC-V spec?

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+            #if CPUINFO_ARCH_RISCV32 || CPUINFO_ARCH_RISCV64
```
prashanthswami, line 60: Do we need this check? This function / this file requires us to be building for RISC-V.

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+    /* Search for ':' on the line. */
+    const char* separator = line_start;
+    for (; separator != line_end; separator++) {
+        if (*separator == ':') {
+            break;
+        }
+    }
+    /* Skip line if no ':' separator was found. */
+    if (separator == line_end) {
+        cpuinfo_log_info("Line %.*s in /proc/cpuinfo is ignored: key/value separator ':' not found",
```
prashanthswami, line 195: Should this be a warning?

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+            break;
+        case 9:
+            if (memcmp(line_start, "processor", key_length) == 0) {
+                const uint32_t new_processor_index = parse_processor_number(value_start, value_end);
+                if (new_processor_index < processor_index) {
+                    /* Strange: decreasing processor number */
+                    cpuinfo_log_warning(
+                        "unexpectedly low processor number %"PRIu32" following processor %"PRIu32" in /proc/cpuinfo",
+                        new_processor_index, processor_index);
+                } else if (new_processor_index > processor_index + 1) {
```
prashanthswami, line 277: Preferably: "new_processor_index > (processor_index + 1)".

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+        case 5:
+            if (memcmp(line_start, "uarch", key_length) == 0) {
+                parse_uarch(value_start, value_end, processor);
+            } else {
+                goto unknown;
+            }
+            break;
+        case 9:
+            if (memcmp(line_start, "processor", key_length) == 0) {
+                const uint32_t new_processor_index = parse_processor_number(value_start, value_end);
```
prashanthswami, line 271: In the case where we continuously fail to read the line and always return 0, I think these conditions will just repeatedly overwrite the state of the '0' processor? We need to also have a condition here where the processor index never progresses. (Or hard-fail when 'parse_processor_number' fails).

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+bool cpuinfo_riscv_linux_parse_proc_cpuinfo(
+        uint32_t max_processors_count,
+        struct cpuinfo_riscv_linux_processor processors[restrict static max_processors_count])
+{
+    struct proc_cpuinfo_parser_state state = {
+            .processor_index = 0,
+            .max_processors_count = max_processors_count,
+            .processors = processors,
+    };
+    cpuinfo_log_debug("");
```
prashanthswami, line 312: Left over from debugging I think.

src/riscv/linux/hwcap.c
```diff
@@ -0,0 +1,37 @@
...
+
+#if CPUINFO_MOCK
+    #include <cpuinfo-mock.h>
+#endif
+
+#if CPUINFO_ARCH_RISCV32 || CPUINFO_ARCH_RISCV64
+    #include <sys/auxv.h>
+#else
+    #define AT_HWCAP 16
+    #define AT_HWCAP2 26
```
prashanthswami, line 11: Do we need HWCAP2? I don't think it does anything on RISC-V. I guess I'm also a bit confused as to why the ARCH here determines how AT_HWCAP is defined - these variable do not seem to be used at all?

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+            }
+            break;
+        case 4:
+            if (memcmp(line_start, "hart", key_length) == 0) {
+                // Do nothing
+            } else {
+                goto unknown;
+            }
+            break;
+        case 5:
```
prashanthswami, line 262: Preferably, let's not handle lines that we don't intend to do anything with - it's just extra code maintenance for us (unless this generates a lot of debug noise?).

src/riscv/linux/hwcap.c
```diff
@@ -0,0 +1,37 @@
...
+        mock_hwcap = hwcap;
+    }
+
+	static uint32_t mock_hwcap2 = 0;
+	void cpuinfo_set_hwcap2(uint32_t hwcap2) {
+		mock_hwcap2 = hwcap2;
+	}
+#endif
+
+#if CPUINFO_ARCH_RISCV32 || CPUINFO_ARCH_RISCV64
```
prashanthswami, line 26: Why is this guarded with ARCH type? BTW I started adding tests for the visionfive board recently: #200 - does this satisfy the hwcap / riscv_hwprobe mocking bits?

src/riscv/linux/init.c
```diff
@@ -1,25 +1,55 @@
...
 #include <cpuinfo/internal-api.h>
 #include <cpuinfo/log.h>
 #include <linux/api.h>
+#include <riscv/api.h>
 #include <riscv/linux/api.h>
 
 /* ISA structure to hold supported extensions. */
 struct cpuinfo_riscv_isa cpuinfo_isa;
 
+static struct cpuinfo_package package;
```
prashanthswami, line 16: Not used I think?

src/riscv/linux/init.c
```diff
@@ -227,75 +257,95 @@ void cpuinfo_riscv_linux_init(void) {
...
+	if (max_present_processors_count != 0) {
+		riscv_linux_processors_count = min(riscv_linux_processors_count, max_present_processors_count);
+		valid_processor_mask |= CPUINFO_LINUX_FLAG_PRESENT;
+	}
+	if (max_possible_processors_count != 0) {
+		riscv_linux_processors_count = min(riscv_linux_processors_count, max_possible_processors_count);
+		valid_processor_mask |= CPUINFO_LINUX_FLAG_POSSIBLE;
+	}
+	if ((max_present_processors_count | max_possible_processors_count) == 0) {
+		cpuinfo_log_error("failed to parse both lists of possible and present processors");
```
prashanthswami, line 288: I'm not quite following why the ARM code is written in the same way - in what case do we need to mix the present and possible processors, given that we can only ever infer data about the present processors?

src/riscv/linux/init.c
```diff
@@ -562,10 +627,26 @@ void cpuinfo_riscv_linux_init(void) {
...
+		/* Populate cache information structures in l1i, l1d */
+		struct cpuinfo_cache temp_l2 = { 0 };
+		cpuinfo_riscv_decode_cache(
+				riscv_linux_processors[processor].core.uarch,
+				&l1i[processor], &l1d[processor], &temp_l2);
+		l1i[processor].processor_start = l1d[processor].processor_start = processor;
+		l1i[processor].processor_count = l1d[processor].processor_count = 1;
+		if (temp_l2.size != 0) {
+			/* Assume L2 is shared by cores in the same cluster */
+			if (riscv_linux_processors[processor].package_leader_id == linux_id) {
```
prashanthswami, line 639: Do you want 'cluster_leader_id' here?

src/riscv/linux/init.c
```diff
@@ -577,6 +658,41 @@ void cpuinfo_riscv_linux_init(void) {
...
+                        .partitions      = temp_l2.partitions,
+                        .line_size       = temp_l2.line_size,
+                        .flags           = temp_l2.flags,
+                        .processor_start = processor,
+                        .processor_count = riscv_linux_processors[processor].package.processor_count,
+                };
+            }
+            processors[processor].cache.l2 = l2 + l2_index;
+        }
+    }
```
prashanthswami, line 694: Couldn't we do this in the loop above, right where we commit the L1 information?

src/riscv/linux/init.c
```diff
@@ -588,8 +704,16 @@ void cpuinfo_riscv_linux_init(void) {
 	cpuinfo_packages_count = valid_packages_count;
 	cpuinfo_uarchs = uarchs;
 	cpuinfo_uarchs_count = valid_uarchs_count;
+	cpuinfo_cache[cpuinfo_cache_level_1i] = l1i;
+	cpuinfo_cache_count[cpuinfo_cache_level_1i] = valid_processors_count;
+	cpuinfo_cache[cpuinfo_cache_level_1d] = l1d;
+	cpuinfo_cache_count[cpuinfo_cache_level_1d] = valid_processors_count;
+	cpuinfo_cache[cpuinfo_cache_level_2]  = l2;
```
prashanthswami, line 711: Extra whitespace.

src/riscv/linux/init.c
```diff
@@ -588,8 +704,16 @@ void cpuinfo_riscv_linux_init(void) {
...
+	cpuinfo_cache_count[cpuinfo_cache_level_2]  = valid_clusters_count;
```
prashanthswami, line 712: Should be 'l2_count'.

src/riscv/linux/riscv-hw.c
```diff
@@ -16,10 +21,6 @@ void cpuinfo_riscv_linux_decode_vendor_uarch_from_hwprobe(
 	};
 	const size_t pairs_count = sizeof(pairs) / sizeof(struct riscv_hwprobe);
 
-	/* In case of failure, report unknown. */
-	*vendor = cpuinfo_vendor_unknown;
-	*uarch = cpuinfo_uarch_unknown;
-
```
prashanthswami, line 22: If we're going to remove this, can we change this function to report a bool success or failure? The caller should know whether or not to trust the information in the structure.

src/riscv/linux/riscv-hw.c
```diff
@@ -1,10 +1,15 @@
...
 #include <sched.h>
-
 #include <cpuinfo/log.h>
+
 #include <riscv/api.h>
 #include <riscv/linux/api.h>
 
+#include <linux/version.h>
+
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6,4,0)
```
prashanthswami, line 9: So . . . .philosophical question to @malfet @GlassOfWhiskey . I know current hardware is not all at 6.4 (I guess, to be more fair, most of it isn't), but hwprobe is supposed to be the "proper" way of querying this information. I think it's fair to say that no one likes the prospect of trying to string-parse a proc-file to infer data.

This actually goes back to my comment on the other file about whether we should take this opportunity to just remove all the features that aren't useful to detect (e.g. why do we want to detect 'm', literally everyone will have 'm'). If we do it that way, then the question becomes "if you're only here to detect, say "V", well V isn't available till Linux 6.5, so you *must* have hwprobe". What do you both think?

prashanthswami (changes requested):

GlassOfWhiskey:
> @GlassOfWhiskey - I sincerely apologize, this was poor GitHub practice on my part. I had written my PR a few months ago on an internal fork and needed it up streamed somewhat quickly, so I pushed it without analyzing the open PRs. When I looked at this issue last, it hadn't been active for a few months, so I was worried it would be inappropriate to ping here and ask for an accelerated landing, just to satisfy my own project timelines. In hindsight, I see that it would have been courteous to message here and coordinate, or CC you on the other PR. I'd like to make sure I don't do this again - should I CC you on any open PRs I have for RISC-V contributions to this repo?

@prashanthswami no problem, it was just a bit of work to merge all the things, but the important thing is to provide community with RISC-V support, no matter how or who :)
I'll start addressing your comments. Since there is a lot of stuff, I'll probably do it incrementally, so that we can track progresses. Then we can squash all commits at the end.

As a general comment @prashanthswami @malfet: if you care so much about the code style and coherence, why don't you or someone in the maintainers team put a `.clang-format` file or something equivalent in the repo? So that contributors can automatically format the code before submitting and you can integrate style adherence directly in the CI pipeline?

prashanthswami:
In regards to style-adherence, I'll defer to @malfet here, I'm only along for the ride :) my format comments were largely just parroting earlier comments Marat / Nikita gave me. Supportive though of the idea, anything that lets my IDE auto-switch to the right format is desirable. Does PyTorch have top-level guidance about what is considered proper code format?

tools/isa-info.c
GlassOfWhiskey: Did it

src/riscv/linux/init.c
GlassOfWhiskey: Corrected

src/riscv/linux/init.c
GlassOfWhiskey: Deleted

src/riscv/linux/api.h
```diff
@@ -1,8 +1,26 @@
...
 
+#if CPUINFO_ARCH_RISCV32 || CPUINFO_ARCH_RISCV64
+    /* arch/riscv/include/uapi/asm/hwcap.h */
+	#define CPUINFO_RISCV_LINUX_FEATURE_A                UINT32_C(0x00000001)
+	#define CPUINFO_RISCV_LINUX_FEATURE_C                UINT32_C(0x00000004)
+	#define CPUINFO_RISCV_LINUX_FEATURE_D                UINT32_C(0x00000008)
+	#define CPUINFO_RISCV_LINUX_FEATURE_F                UINT32_C(0x00000020)
+	#define CPUINFO_RISCV_LINUX_FEATURE_I                UINT32_C(0x00000100)
+	#define CPUINFO_RISCV_LINUX_FEATURE_M                UINT32_C(0x00001000)
+	#define CPUINFO_RISCV_LINUX_FEATURE_V                UINT32_C(0x00200000)
```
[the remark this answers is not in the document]
GlassOfWhiskey: Ordered

include/cpuinfo.h
```diff
@@ -76,6 +78,14 @@
 	#define CPUINFO_ARCH_PPC64 0
 #endif
 
+#ifndef CPUINFO_ARCH_RISCV32
```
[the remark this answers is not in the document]
GlassOfWhiskey: Removed

src/riscv/linux/cpuinfo.c
GlassOfWhiskey: Fixed

src/riscv/linux/cpuinfo.c
GlassOfWhiskey: Sorted

src/riscv/linux/init.c
GlassOfWhiskey: Removed

src/riscv/linux/cpuinfo.c
GlassOfWhiskey: Removed checks, after verifying it still compiles on non-RISCV archs :)

CMakeLists.txt
GlassOfWhiskey: Removed whitespace and question mark

CMakeLists.txt
GlassOfWhiskey: Removed whitespace

src/riscv/linux/riscv-hw.c
markdryan: This #ifdef does not fix the RISC-V build on machines with a 6.4 kernel or greater.  When compiling this branch on a RISC-V 64 Ubuntu 23.10 VM, which has a 6.5 kernel, I get the following errors

```
cpuinfo/src/riscv/linux/riscv-hw.c:11:10: fatal error: sys/hwprobe.h: No such file or directory
   11 | #include <sys/hwprobe.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```
It's not clear to me where sys/hwprobe.h is supposed to come from.  @prashanthswami I think you added this originally.  Were you testing with your own build of glibc?

Also, IIUC, with this compile time check on the kernel version, hwprobe will only be used if both the build machine and the system running the binary have a 6.4 kernel or greater.  If we build on Ubuntu 22.04 for example but run on 23.10 the hwprobe path would not be taken here even though 23.10 has a 6.5 kernel with hwprobe support.

Other projects that are using hwprobe right now get around this problem by making the syscall directly. [OpenJDK](https://github.com/openjdk/jdk/blob/master/src/hotspot/os_cpu/linux_riscv/riscv_hwprobe.cpp) duplicates the constants from the kernel headers needed for the sys call and then does a direct syscall. It doesn't include the kernel header files at all.  A nicer solution would be to first check to see if the <asm/hwprobe.h> exists either using a build system check that defines a macro if a program that includes <asm/hwprobe.h> builds or perhaps using the trick the glibc patch uses

```
#ifdef __has_include
# if __has_include (<asm/hwprobe.h>)
#  include <asm/hwprobe.h>
# endif
#endif
```

Regardless of the method used, if the include file is not found,  manually defined constants for the hwprobe syscall number and the various bitmasks can be defined, so the syscall can still be issued, regardless of the kernel version on the build machine.

src/riscv/linux/riscv-isa.c
```diff
@@ -1,44 +1,33 @@
...
-#define COMPAT_HWCAP_ISA_F	(1 << ('F' - 'A'))
-#define COMPAT_HWCAP_ISA_D	(1 << ('D' - 'A'))
-#define COMPAT_HWCAP_ISA_C	(1 << ('C' - 'A'))
-#define COMPAT_HWCAP_ISA_V	(1 << ('V' - 'A'))
-
 void cpuinfo_riscv_linux_decode_isa_from_hwcap(
 		struct cpuinfo_riscv_isa isa[restrict static 1]) {
-	const unsigned long hwcap = getauxval(AT_HWCAP);
+	unsigned long hwcap = 0;
+	cpuinfo_riscv_linux_hwcap_from_getauxval(&hwcap);
```
markdryan, line 10: I get a compiler warning on this line

```
warning: passing argument 1 of ‘cpuinfo_riscv_linux_hwcap_from_getauxval’ from incompatible pointer ty
pe [-Wincompatible-pointer-types]                                                                                                                                     
   10 |         cpuinfo_riscv_linux_hwcap_from_getauxval(&hwcap);                                                                                                     
      |                                                  ^~~~~~                                                                                                       
      |                                                  |
      |                                                  long unsigned int *
```

src/riscv/linux/riscv-hw.c
prashanthswami: My primary use-case was [Android on RISC-V](https://github.com/google/android-riscv64), so I likely overcorrected here by testing primarily against the Android NDK (note, there's no official NDK with RISC-V support yet, but the public canary builds work). This header is available in [bionic](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/kernel/uapi/asm-riscv/asm/hwprobe.h), so building against the NDK does not complain.

The proposal to check for the header explicitly seems reasonable enough to me and there's precedent for it.

As a side note - we should indeed have a CI bot for this. I can't add one for Android just yet, because the common Github workflows don't support canary Android NDKs, but I've filed #206 to do so when it is. Separately, it would probably be valuable to run a linux-riscv64 build to catch stuff like this, so I'll figure out that out.

markdryan: Ah, that explains it.  It looks like you're pulling in this [header](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/sys/hwprobe.h) which contains a definition of __riscv_hwprobe that appears to be compatible with the pending glibc patch.  Ultimately, the thing to do would be to check for <sys/hwprobe.h> and, if it's not present, fall back to code that makes the syscall directly.  However, perhaps until the glibc patch is merged, <sys/hwprobe.h> should only be included when building for Android.  A manual riscv_hwprobe syscall could then be used for other Linux distributions for the time being.

src/riscv/linux/init.c
```diff
@@ -509,8 +571,9 @@ void cpuinfo_riscv_linux_init(void) {
 	size_t valid_clusters_index = 0;
 	size_t valid_packages_index = 0;
 	size_t valid_uarchs_index = 0;
+	size_t l2_count = 0;
 	last_uarch = cpuinfo_uarch_unknown;
-	for (size_t processor = 0; processor < max_processor_id; processor++) {
+	for (size_t processor = 0; processor < riscv_linux_processors_count; processor++) {
 		if (!bitmask_all(riscv_linux_processors[processor].flags, CPUINFO_LINUX_FLAG_VALID)) {
```
markdryan, line 577: Does the qsort above guarantee that all the invalid processors are at the end of riscv_linux_processors?  If so, should this be a break rather than a continue?

markdryan (commented):
A couple more comments about the current use of riscv_hwprobe.  The use of hwprobe I'm commenting on here was introduced in an earlier [PR](https://github.com/pytorch/cpuinfo/pull/190) but as we're having the conversation about hwprobe here I thought it worth mentioning.  Happy to copy these comments to an issue if this is preferred.

1. We're calling hwprobe once for each processor to retrieve the vendor id and uarch.  If we have lots of processors this will result in lots of syscalls (and lots of failures on kernels older than 6.4).  There's a good chance that all the processors will have the same vendor and uarch, so it's probably worth trying to retrieve the vendor and uarch for all the processors we're interested in a single syscall, outside of a loop.  If -1 is returned for either value, different processors do have different values for uarch and vendor id and so they would need to be queried on a processor by processor basis as is currently done.
2. We don't use hwprobe to detect the ISA extensions.  We're only using hwcap, which cannot detect multi-letter extensions.  It would be better to check for extensions with hwprobe first, falling back to hwcap if hwprobe is not available.  This could be done in the same call used to retrieve the vendor and uarch for all processors as recommend in point 1 above.  Using risc_hwprobe in this way would allow us to detect Zba, Zbb and Zbs (and more extensions in the future).

src/riscv/cache.c
```diff
@@ -0,0 +1,52 @@
...
+#include <cpuinfo.h>
+#include <cpuinfo/internal-api.h>
+#include <cpuinfo/log.h>
+
+void cpuinfo_riscv_decode_cache(
+        enum cpuinfo_uarch uarch,
+        struct cpuinfo_cache l1i[restrict static 1],
+        struct cpuinfo_cache l1d[restrict static 1],
+        struct cpuinfo_cache l2[restrict static 1])
+{
```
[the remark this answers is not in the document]
GlassOfWhiskey: Fixed

src/riscv/cache.c
```diff
@@ -0,0 +1,52 @@
...
+    l1i->partitions = 1;
+    l1d->sets = l1d->size / (l1d->associativity * l1d->line_size);
+    l1d->partitions = 1;
+    if (l2->size != 0) {
+        l2->sets = l2->size / (l2->associativity * l2->line_size);
+        l2->partitions = 1;
+    }
+}
+
+uint32_t cpuinfo_riscv_compute_max_cache_size(const struct cpuinfo_processor* processor) {
```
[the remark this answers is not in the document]
GlassOfWhiskey: Fixed

src/riscv/linux/cpuinfo.c
GlassOfWhiskey: Removed

src/riscv/linux/cpuinfo.c
GlassOfWhiskey: Changed to `cpuinfo_log_warning`

src/riscv/linux/cpuinfo.c
GlassOfWhiskey: Added condition

src/riscv/cache.c
```diff
@@ -0,0 +1,52 @@
...
+#include <cpuinfo/log.h>
+
+void cpuinfo_riscv_decode_cache(
+        enum cpuinfo_uarch uarch,
+        struct cpuinfo_cache l1i[restrict static 1],
+        struct cpuinfo_cache l1d[restrict static 1],
+        struct cpuinfo_cache l2[restrict static 1])
+{
+    switch(uarch) {
+        case cpuinfo_uarch_u74_mc:
```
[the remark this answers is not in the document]
GlassOfWhiskey: Added reference to SiFive manual

src/riscv/linux/cpuinfo.c
```diff
@@ -0,0 +1,315 @@
...
+
+static uint32_t parse_processor_number(
+        const char* processor_start,
+        const char* processor_end)
+{
+    const size_t processor_length = (size_t) (processor_end - processor_start);
+
+    if (processor_length == 0) {
+        cpuinfo_log_warning("Processor number in /proc/cpuinfo is ignored: string is empty");
+        return 0;
```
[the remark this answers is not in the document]
GlassOfWhiskey: Changed `0` with `-1`

src/riscv/linux/api.h
```diff
@@ -65,5 +86,16 @@ CPUINFO_INTERNAL void cpuinfo_riscv_linux_decode_vendor_uarch_from_hwprobe(
 	enum cpuinfo_vendor vendor[restrict static 1],
 	enum cpuinfo_uarch uarch[restrict static 1]);
 
+CPUINFO_INTERNAL void cpuinfo_riscv_linux_count_cluster_processors(
+	uint32_t max_processors,
+	struct cpuinfo_riscv_linux_processor processors[restrict static max_processors]);
```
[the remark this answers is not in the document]
GlassOfWhiskey: Removed and commented other functions

src/riscv/linux/init.c
GlassOfWhiskey: Fixed

src/riscv/linux/init.c
GlassOfWhiskey: No because we don't know the value of `l2_count`. THe program first computes how many `l2` caches are present, then allocates the `l2` data structure and populates it. We can change the logic but it will behave diffferently from what it does now.

src/riscv/cache.c
```diff
@@ -0,0 +1,52 @@
...
+               .line_size = 64
+           };
+           *l2 = (struct cpuinfo_cache) {
+                .size = 2 * 1024 * 1024,
+                .associativity = 16,
+                .line_size = 64
+           };
+           break;
+        default:
+            cpuinfo_log_warning("target uarch not recognized; cache data is not populated");
```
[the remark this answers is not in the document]
GlassOfWhiskey: Implemented both bool return and early return, as suggested.

src/riscv/linux/riscv-hw.c
markdryan: @malfet, @GlassOfWhiskey , @prashanthswami The build issues on mainline Linux distributions are still not addressed by this patch.  It hasn't been possible to build pytorch/cpuinfo on Ubuntu 23.10 for over 3 weeks and this patch in it's current form won't help.  Would there by any objections from anyone if I submit a PR to fix the build issues?  The PR would only address the build issues to avoid introducing too many conflicts.  Other issues with hwprobe, like the fact we're calling it far too often in most cases, could be addressed in a subsequent PR.

GlassOfWhiskey: @markdryan no problem from my side. Also because probably things will go a bit longer in approving this PR as many aspects are involved.
I missed what was the preferred solution among the ones proposed so I didn't do anything on that direction for now. Feel free to proceed as you prefer.
The `__has_include` path is probably the most elegant one. It is not fully C standard, but it is now part of the C++ standard since C++17 so I'd vote for that one.

include/cpuinfo.h
```diff
@@ -1933,14 +1942,6 @@ static inline bool cpuinfo_has_riscv_i(void) {
 	#endif
 }
 
-static inline bool cpuinfo_has_riscv_e(void) {
-	#if CPUINFO_ARCH_RISCV32
-		return cpuinfo_isa.e;
-	#else
-		return false;
-	#endif
-}
```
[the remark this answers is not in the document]
fbarchard: I'd like to see everything removed that is not immediately useful.  Attempting to run without 'base isa' is not useful.

fbarchard (approved):

PR #152, opened 2023-06-21
Fill frequency field on arm linux

blaztinn:
Copied the linux arm implementation value of the max frequency to public `cpuinfo_core.frequency` and `cpuinfo_cluster.frequency`.

I'm not sure if `cpuinfo_*.frequency` is meant to represent the current or max frequency. It is not apperent to me from the struct member comments. This patch assumes max is meant.


Maratyszcza:
Needs tests

PR #153, opened 2023-07-03
cmake: add USE_SYSTEM_* options

ConnorBaker:
Similar to the work done in https://github.com/pytorch/pytorch/pull/37137, this adds the following CMake options:

- `USE_SYSTEM_LIBS`
- `USE_SYSTEM_GOOGLEBENCHMARK`
- `USE_SYSTEM_GOOGLETEST`

This is particularly useful in the context of Nix, where we can build these libraries once and then re-use them elsewhere to avoid rebuilding vendors dependencies.


ConnorBaker:
@malfet @Maratyszcza would either of you have the bandwidth to take a look at this?

malfet (approved):

malfet:
@ConnorBaker Looks good to me, though I wonder if `USE_SYSTEM_${LIBNANE}` option is at all needed. I.e. what would be the realistic scenario, when one wants to use `gtest` from OS, but `gmock` from third_party folder.

CMakeLists.txt
```diff
@@ -23,6 +23,9 @@ OPTION(CPUINFO_BUILD_UNIT_TESTS "Build cpuinfo unit tests" ON)
 OPTION(CPUINFO_BUILD_MOCK_TESTS "Build cpuinfo mock tests" ON)
 OPTION(CPUINFO_BUILD_BENCHMARKS "Build cpuinfo micro-benchmarks" ON)
 OPTION(CPUINFO_BUILD_PKG_CONFIG "Build pkg-config manifest" ON)
+OPTION(USE_SYSTEM_LIBS "Use system libraries instead of downloading and building them" OFF)
+OPTION(USE_SYSTEM_GOOGLE_BENCHMARK "Use system Google Benchmark library instead of downloading and building it" ${USE_SYSTEM_LIBS})
```
Maratyszcza, line 27: Please use `USE_SYSTEM_GOOGLE_BENCHMARK` for consistency with `GOOGLEBENCHMARK_SOURCE_DIR`

CMakeLists.txt
```diff
@@ -23,6 +23,9 @@ OPTION(CPUINFO_BUILD_UNIT_TESTS "Build cpuinfo unit tests" ON)
...
+OPTION(USE_SYSTEM_GOOGLE_TEST "Use system Google Test library instead of downloading and building it" ${USE_SYSTEM_LIBS})
```
Maratyszcza, line 28: Ditto, change to `USE_SYSTEM_GOOGLETEST`

deps/clog/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ ELSE()
   OPTION(CLOG_LOG_TO_STDIO "Log errors, warnings, and information to stdout/stderr" ON)
 ENDIF()
 OPTION(CLOG_BUILD_TESTS "Build clog tests" ON)
+OPTION(USE_SYSTEM_LIBS "Use system libraries instead of downloading and building them" OFF)
+OPTION(USE_SYSTEM_GOOGLE_TEST "Use system Google Test library instead of downloading and building it" ${USE_SYSTEM_LIBS})
```
Maratyszcza, line 18: Ditto

CMakeLists.txt
malfet: @Maratyszcza I guess you've meant `USE_SYSTEM_GOOGLEBENCHMARK`? ;)

ConnorBaker:
Thank you for the feedback! I've addressed comments and force-pushed the changes.

@malfet I think it makes sense to have `USE_SYSTEM_LIBS` to avoid needing to set multiple flags. With respect to the question about `gmock`: I thought `gmock` was brought in by `find_package(GTest)`? Or was this a question of how to support system `gtest` but third-party `gmock`?

@Maratyszcza are the names good now? Also (because I have multiple PRs open doing the same thing), would you recommend using that naming instead of `GOOGLE_*`?

malfet (approved):

Maratyszcza:
@Maratyszcza I don't have a preference for `GOOGLETEST` vs `GOOGLE_TEST`, so long as the naming is consistent

PR #156, opened 2023-07-12
Fix detection of CPU vendor on ARM

Maratyszcza:
- `cpuinfo_vendor` is the vendor of the CPU core, not CPU package
- Remove `cpuinfo_vendor_ampere` as cpuinfo currently doesn't detect any cores designed by Ampere Computing

### DIFF
--- a/include/cpuinfo.h
+++ b/include/cpuinfo.h
@@ -188,8 +188,6 @@ enum cpuinfo_vendor {
 	 * Processors are variants of AMD cores.
 	 */
 	cpuinfo_vendor_hygon    = 16,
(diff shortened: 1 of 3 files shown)


Maratyszcza:
@malfet PTAL

malfet (approved):
Sure, though cc: @Blackhex who added those

Blackhex:
@Maratyszcza I've added only "Ampere", the "Qualcomm" was already there... Nevertheless, the new way will make things easier, thanks for it. Did you tested it on Ampere Altra, Microsoft Surface X and Microsoft Dev Kit 2023 already or should we test it?

PR #157, opened 2023-07-13
Fix potential uninitialized variable UB

Maratyszcza:
Fix potential undefined behavior when parsing /proc/cpuinfo Hardware string


Maratyszcza:
@malfet PTAL

malfet (approved):

PR #167, opened 2023-07-20
Added Apple Silicon support to WoA

vladimir-aubrecht:
Added support for Apple Silicon.

### DIFF
--- a/src/arm/windows/init.c
+++ b/src/arm/windows/init.c
@@ -94,6 +94,23 @@ static struct woa_chip_info woa_chips[] = {
 				3000000000
 			}
 		}
+	},
+	/* Apple Silicon */
+	{
+		L"Apple Silicon",
+		woa_chip_name_apple_silicon,
+		{
+			{
+				cpuinfo_vendor_apple,
+				cpuinfo_uarch_firestorm,
+				3228000000
+			},
+			{
+				cpuinfo_vendor_apple,
+				cpuinfo_uarch_icestorm,
+				2064000000
+			}
+		}
 	}
 };
 
--- a/src/arm/windows/windows-arm-init.h
+++ b/src/arm/windows/windows-arm-init.h
@@ -7,7 +7,8 @@ enum woa_chip_name {
 	woa_chip_name_microsoft_sq_3 = 2,
 	woa_chip_name_ampere_altra = 3,
 	woa_chip_name_unknown = 4,
-	woa_chip_name_last = woa_chip_name_unknown
+	woa_chip_name_apple_silicon = 5,
+	woa_chip_name_last = woa_chip_name_apple_silicon
 };
 
 /* Topology information hard-coded by SoC/chip name */


Maratyszcza:
How do you differentiate M1/M1 Pro/M2/etc?

vladimir-aubrecht:
> How do you differentiate M1/M1 Pro/M2/etc?

Great question, not so great answer ... 😕

I tried to preserve compatibility with current code and from that perspective all M1s are identical - they have same frequency and same types of high/low efficiency cores. This is not optimal, but if I did not miss something, returning more information - like amount of cores - would require breaking changes and this change at least prevents failures on M1s.

About distinguishing between M1 & M2, honestly I don't know. I don't have M2 to test it and did not find what string will be shown in Windows Registry for M2. I was hoping for different string or somebody updating it when it will become problem.

Maratyszcza:
The current solution would likely produce wrong results on M2, which is a big no for cpuinfo. I'd rather have it report `unknown` that report wrong data. Please borrow an M2 Mac and try to figure out the answer. I found that `SYSTEM_INFO::wProcessorLevel` returned by `GetSystemInfo` seems to contain the "Part" component of the MIDR. Maybe this would be sufficient to distinguish M1, M2, and later cores.

vladimir-aubrecht:
@Maratyszcza I checked ``SYSTEM_INFO::wProcessorLevel`` and unfortunately it has value 0 on both M1 and M2. This could be due to running in VM as that is only way how to run Windows on M1 & M2 at this moment.

I tried to check also other attributes, but as far as I am aware of, there is no way how to distinguish M1 & M2 at this moment 😕 (you can check attached screenshots of Windows Registry + output of ``wmic cpu get`` which provides same value as ``SYSTEM_INFO::wProcessorLevel`` in Level field and more ...).

I am not sure how to follow up on this due to inability to distinguish between those CPUs. Do we need distinction between M1 & M2? I mean they are both Apple Silicon, so reporting Apple Silicon would be technically correct in all cases and it's more question of if we need distinction between them for something? What do you think?

Registry M1
Registry M2

[WMIC M1](https://github.com/pytorch/cpuinfo/files/12184502/m1.txt)
[WMIC M2](https://github.com/pytorch/cpuinfo/files/12184504/m2.txt)

Maratyszcza:
Without the ability to distinguish variants of Apple Silicon all we can do is set `vendor` to `cpuinfo_vendor_apple`, as all other information is unreliable. Narrowing down this change to just the `vendor` setting would be acceptable, although not very useful.

I'd like to encourage you to check within Microsoft what other ways there can be to retrieve information about the cores. In particular, if you could get MIDR data for the cores, we could decode the microarchitecture and use them for ISA feature detection (e.g. Apple M2 supports BF16 and I8MM while Apple M1 doesn't).

Maratyszcza:
One hint might be the CP4020...CP5801 values reported in the registry. They are evidently different across M1 and M2, but it is unclear what they represent. CP4020 looks like MIDR, albeit a fake one, probably others are different system registers.

Maratyszcza:
[This source snippet](https://github.com/ziglang/zig/blob/0461a64a93f0596e98b62d596bb547e5455577d2/lib/std/zig/system/windows.zig#L243-L255) suggest that CP NNN these are indeed system registers.

PR #170, opened 2023-07-27
Add support for linux ppc64le architecture

npanpaliya:
This change is needed for cpuinfo to build on linux ppc64le architecture. Without this change, Tensorflow's build fails on ppc64le.


npanpaliya:
> This doesn't actually add support for Linux/PPC64, only (probably) makes cpuinfo build on this platform

Yes @Maratyszcza ! Yes, this is just to get cpuinfo built as part of Tensorflow build. If you could help me on where and how can we add full support of Linux ppc64le in cpuinfo code, it will be great.

smuzaffar:
Using this patch I was able to build Tensorflow 2.15.0 for pc64le.

PR #175, opened 2023-08-09
Fix various bugs in Windows on Arm support

LDong-Arm:


src/arm/windows/init.c
```diff
@@ -137,7 +134,7 @@ bool get_core_uarch_for_efficiency(
 static bool read_registry(
```
Maratyszcza, line 134: While we're at it, make `read_registry` return `wchat_t*` with the text buffer (return `NULL` on error).

Maratyszcza (changes requested):
Thank you for the fixes! One minor request, and its ready to merge

src/arm/windows/init.c
LDong-Arm: Thanks for the suggestion, I've added a new commit on top.

PR #179, opened 2023-08-16
Workaround unimplemented GetMaximumProcessorGroupCount on WINE

Maratyszcza:


mrpippy:
#135 removed this workaround, why is it needed again? Are really old versions of Wine (like earlier than 6.7) being used?

PR #197, opened 2023-11-02
[arm] fix the logic for identifying the valid processors

snadampal:
The current logic for valid processor detection is reporting all cpus irrespective of whether they are online or not. so, it's causing thread over-subscription for the scenarios where the online cpu count < the actual cpus. This is fixed by publishing only the online cpu count as the valid processors.


snadampal:
appreciate if someone can review this PR?

malfet (approved):

PR #199, opened 2023-11-14
Fix size check of max processor count

prashanthswami:
On 64-bit systems, size_t will not overflow when the function to get max processors returns UINT32_MAX. Explicitly check for the proper value so that both 32-bit and 64-bit systems will succeed.

### DIFF
--- a/src/riscv/linux/init.c
+++ b/src/riscv/linux/init.c
@@ -244,7 +244,7 @@ void cpuinfo_riscv_linux_init(void) {
 	 * processors. It is not a count of the number of processors on the
 	 * system.
 	 */
-	const size_t max_processor_id = 1 +
+	const uint32_t max_processor_id = 1 +
 		cpuinfo_linux_get_max_present_processor(
 				cpuinfo_linux_get_max_processors_count());
 	if (max_processor_id == 0) {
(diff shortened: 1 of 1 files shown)


prashanthswami:
@malfet Flagging as the fix for the size check you flagged in the recently landed RISC-V CL.

src/riscv/linux/init.c
```diff
@@ -247,7 +247,7 @@ void cpuinfo_riscv_linux_init(void) {
 	const size_t max_processor_id = 1 +
```
malfet, line 247: Just curious, what is the downside of casting this to `uint32_t`? and this way you don't need to cast that will throw compiler warning on 32-bit platform?

Also, I don't know much about RISC-V, is this a defined behavior the plafrom/compiler it overflows to 0?
If I'm not mistaken, C standard does not define how system should behave in this case

prashanthswami: I could've sworn there was a concrete reason I didn't do this, but it's been too long since I originally wrote this :/ - however, trying this out right now, I don't see any obvious harm and it seems to work. I'll put up that change shortly.

In regards to the overflow, I took a brief look into what the deal was here - it turns out, C **does** define the system behavior for integer overflow, but only when that integer is *unsigned**! The specific text is:

"A computation involving unsigned operands can never overﬂow, because a result that cannot be represented by the resulting unsigned integer type is reduced modulo to the number that is one greater than the largest value that can be represented by the resulting type."

There are some nice StackOverflow posts about this too, but the tl;dr if it's unsigned, it's guaranteed to wrap-around to 0 on overflow.

malfet (approved):

PR #200, opened 2023-11-17
Add starfive-visionfive-v2 as example rv64 board

prashanthswami:
We should ideally have at least one RISC-V board in our test group to limit future regression. Further work will have to be done to actually enable a toolchain that can build this test. The Starfive VisionFive V2 board was chosen here as an open-source board that has been referenced in the public issues.

Confirmed success of this mocked device test on a RISC-V QEMU instance.


src/riscv/linux/riscv-hw.c
```diff
@@ -26,9 +37,13 @@ void cpuinfo_riscv_linux_decode_vendor_uarch_from_hwprobe(
...
 
 	/* Request all available information from hwprobe. */
-	int ret = __riscv_hwprobe(pairs, pairs_count,
-                                  cpu_set_size, (unsigned long*)cpu_set,
-                                  0 /* flags */);
+#if CPUINFO_MOCK
+	int ret = g_riscv_hwprobe(pairs, pairs_count, cpu_set_size,
+				  (unsigned long*)cpu_set, 0 /* flags */);
+#else
+	int ret = __riscv_hwprobe(pairs, pairs_count, cpu_set_size,
```
markdryan, line 44: The cpu_set_size parameter is incorrect here.  riscv_hwprobe is expecting the size in bytes of the cpu_set and not the number of cpus, e.g., CPU_ALLOC_SIZE(cpus_set_size).

Also, the use of CPU_SET doesn't look quite right.  I realise this is code was added in a previous PR, but as you're modifying it perhaps it could be updated here.

1. There's no call to CPU_ZERO_S to zero out the allocated set.
2. I think you need to use CPU_SET_S with dynamically allocate cpu_sets rather than CPU_SET, e.g., CPU_SET_S(processor, CPU_ALLOC_SIZE(cpus_set_size), cpu_set)
3. There's no call to CPU_FREE so I guess the cpu_set is leaked.

prashanthswami: Ouch, right on all points. Honestly, I'm a little confused how missing all of those things seems to produce a correct result on my local device . . . anyways, because these issues are a bit far removed from the intent of this patch set, I'll spin a separate one to tackle this and CC you.

prashanthswami: Actually, quick side question: Digging into why I made the mistake on the cpu_set_size, I'm surprised that the [kernel interface](https://docs.kernel.org/arch/riscv/hwprobe.html) says:

"Usermode can supply NULL for cpus and 0 for cpu_count as a shortcut for all online CPUs." and calls this variable 'cpu_count' - both heavily imply that it's looking for the number of elements in the set, and not the set size. However, it's pretty clear looking at the [actual implementation](https://github.com/torvalds/linux/blob/c2d5304e6c648ebcf653bace7e51e0e6742e46c8/arch/riscv/kernel/sys_riscv.c#L271) that they indeed are trying to compute the size in bytes. Out of curiosity, it it just always assumed that when operating on CPU_SETs, that they mean CPU_ALLOC_SIZE? Otherwise, would it be reasonable to poke at upstream Linux forums and ask that this be labeled something more reasonable?

prashanthswami: To tie these two issues together, the fix is available here for your review: https://github.com/pytorch/cpuinfo/pull/205

I'll rebase this CL and likely block it on the other one landing.

markdryan: > "Usermode can supply NULL for cpus and 0 for cpu_count as a shortcut for all online CPUs." and calls this variable 'cpu_count' - both heavily imply that it's looking for the number of elements in the set, and not the set size. However, it's pretty clear looking at the [actual implementation](https://github.com/torvalds/linux/blob/c2d5304e6c648ebcf653bace7e51e0e6742e46c8/arch/riscv/kernel/sys_riscv.c#L271) that they indeed are trying to compute the size in bytes. Out of curiosity, it it just always assumed that when operating on CPU_SETs, that they mean CPU_ALLOC_SIZE? Otherwise, would it be reasonable to poke at upstream Linux forums and ask that this be labeled something more reasonable?

It is a bit confusing.  I only figured this out myself by looking at the kernel sources.  The good news is that there is a [patch](https://lore.kernel.org/all/20231020130515.424577-9-ajones@ventanamicro.com/) to make the API and documentation less confusing.

prashanthswami: While this is now addressed, I'll likely wait on #148 - this CL is likely going to conflict with the CPUINFO_MOCK stuff added there.

PR #204, opened 2023-11-21
Add .clang-format to enforce project style

prashanthswami:
The settings here match the current settings for the pytorch/pytorch project, with the exception that 8-character-width tabs are preferred in place of spaces.

In addition to adding .clang-format, we also run it once on the whole code-base to clean up the existing usages.

Review: @malfet 
CC: @GlassOfWhiskey


prashanthswami:
Are we opinionated about the clang-format version? By default, the Github action tracks to version 13, but we could set it higher.

.clang-format
```diff
@@ -0,0 +1,85 @@
...
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Attach
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: false
+ColumnLimit:     80
```
malfet, line 38: Should we set it to 120?

prashanthswami: Done!

malfet (approved):

Issue #124, opened 2022-12-23
RISC-V support


kassane:
Especially RISC-V64.

There is some expectation of adding support for the new RISCV architecture, even though it is an emulated environment for example QEMU.

VictorElHajj:
:+1: This has been an issue for myself as well

petterreinholdtsen:
According to https://bugs.debian.org/1016623, /proc/cpuinfo can look like this:

```
processor	: 0
hart		: 2
isa		: rv64imafdc
mmu		: sv39
uarch		: sifive,bullet0

processor	: 1
hart		: 1
isa		: rv64imafdc
mmu		: sv39
uarch		: sifive,bullet0

processor	: 2
hart		: 3
isa		: rv64imafdc
mmu		: sv39
uarch		: sifive,bullet0

processor	: 3
hart		: 4
isa		: rv64imafdc
mmu		: sv39
uarch		: sifive,bullet0
```

leuc:
```
cat /proc/device-tree/model
StarFive VisionFive V2

cat /proc/device-tree/compatible
starfive,visionfive-v2starfive,jh7110

cat /proc/device-tree/serial-number
VF7110A1-2250-D008E000-00000824

cat /proc/cpuinfo
processor	: 0
hart		: 1
isa		: rv64imafdc
mmu		: sv39
uarch		: sifive,u74-mc

processor	: 1
hart		: 2
isa		: rv64imafdc
mmu		: sv39
uarch		: sifive,u74-mc

processor	: 2
hart		: 3
isa		: rv64imafdc
mmu		: sv39
uarch		: sifive,u74-mc

processor	: 3
hart		: 4
isa		: rv64imafdc
mmu		: sv39
uarch		: sifive,u74-mc
```

enh-google:
https://github.com/pytorch/cpuinfo/pull/190

PR #205, opened 2023-11-22
Fix CPU_SET dynamic allocation and leak

prashanthswami:
The initial implementation had a number of issues:
- The allocation of the CPU_SET should be checked for a NULL return.
- The CPU_*_S macros should be used when working with dynamic sets.
- The CPU_SET needs to be cleared via CPU_ZERO_S before use.
- Dynamic CPU_SETs need to be freed after use.
- The __riscv_hwprobe syscall is expecting a set *size* not a *count*.

Reviewer: @malfet 
CC: @GlassOfWhiskey @markdryan

### DIFF
--- a/src/riscv/linux/riscv-hw.c
+++ b/src/riscv/linux/riscv-hw.c
@@ -21,9 +21,16 @@ void cpuinfo_riscv_linux_decode_vendor_uarch_from_hwprobe(
 	*uarch = cpuinfo_uarch_unknown;
 
 	/* Create a CPU set with this processor flagged. */
-	const size_t cpu_set_size = processor + 1;
-	cpu_set_t* cpu_set = CPU_ALLOC(cpu_set_size);
-	CPU_SET(processor, cpu_set);
+	const size_t cpu_count = processor + 1;
+	cpu_set_t* cpu_set = CPU_ALLOC(cpu_count);
+	if (cpu_set == NULL) {
+		cpuinfo_log_warning("failed to allocate space for cpu_set");
+		return;
+	}
+
+	const size_t cpu_set_size = CPU_ALLOC_SIZE(cpu_count);
+	CPU_ZERO_S(cpu_set_size, cpu_set);
+	CPU_SET_S(processor, cpu_set_size, cpu_set);
 
 	/* Request all available information from hwprobe. */
 	int ret = __riscv_hwprobe(pairs, pairs_count,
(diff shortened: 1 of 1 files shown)


markdryan:
@prashanthswami I tried building cpuinfo today an a RISC-V VM (Ubuntu 23.10 with glibc 2.38) and the build failed with the following error

```
cpuinfo/src/riscv/linux/riscv-hw.c:1:10: fatal error: sys/hwprobe.h: No such file or directory
    1 | #include <sys/hwprobe.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

This was the main branch without this PR, so it looks like the RISC-V builds are currently broken (they don't seem to be tested by the CI).   It looks to me like you're relying on glibc's implementation of hwprobe, but AFAIK this isn't merged yet.  How did you get it to build?  Did you build your own glibc with the hwprobe patch applied?  In any case, we can't really assume that this header will always be available, not for a good few years at least.

Other projects that are using hwprobe right now are making the syscall directly, which is probably the safest thing to do for now.  [OpenJDK](https://github.com/openjdk/jdk/blob/master/src/hotspot/os_cpu/linux_riscv/riscv_hwprobe.cpp) duplicates the constants from the kernel headers needed for the sys call and then does a direct syscall.  A nicer solution would be to first check to see if the <asm/hwprobe.h> exists either using a build system check or perhaps using

```
#ifdef __has_include
# if __has_include (<asm/hwprobe.h>)
#  include <asm/hwprobe.h>
# endif
#endif
```

and manually defining the constants if asm/hwprobe is not available before making the syscall.

GlassOfWhiskey:
What I was trying to do in #148 is to check the kernel version at compile time.
If you prefer, I can do a different check, or move it directly at runtime. Please let me know what you prefer, but please do not split again the discussion over multiple PRs otherwise it becomes impossible to follow the flow.

markdryan:
> What I was trying to do in #148 is to check the kernel version at compile time. If you prefer, I can do a different check, or move it directly at runtime. Please let me know what you prefer, but please do not split again the discussion over multiple PRs otherwise it becomes impossible to follow the flow.

Where would you like to have the discussions?  Over at #148 or perhaps #124?

prashanthswami:
+1 Let's keep discussions about swapping hwprobe includes over in #148 - this PR is purely to resolve the CPU_SET usage error.

PR #207, opened 2023-11-29
Allow specifying GTest/benchmark locations with -D${packagename}_ROOT

iskunk:
I have my own builds of the **GTest** and **benchmark** libraries that I want to use in the cpuinfo build, but the `USE_SYSTEM_LIBS` option assumes that these are installed in the default compiler/linker paths, without allowing for an alternate location (e.g. somewhere under `$HOME` or `/opt`).

This patch enables the standard `*_ROOT` convention supported by `find_package()` for these libraries, i.e. `-DGTest_ROOT=/path/to/gtest-prefix` and `-Dbenchmark_ROOT=/path/to/benchmark-prefix`.


iskunk:
Updated after confirming that the source-download path can also link against `GTest::gtest` and the like, thus rendering unnecessary the `*_LIB` variables.

iskunk:
Also confirmed that the `INCLUDE_DIRECTORIES()` directives are not needed.

PR #208, opened 2023-11-29
Free resources in cpuinfo_deinitialize

### LINKED ISSUE #150: Missing cpuinfo_deinitialize implementation
When using ONNX Runtime (which has PyTorch cpuinfo as a static dependency) with a Linux program built with [LeakSanitizer](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dleak)/[AddressSanitizer](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003daddress), memory leaks are reported:

```
=================================================================
==11753==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 576 byte(s) in 1 object(s) allocated from:
    #0 0x2b11dd429917 in __interceptor_calloc (/path/to/libasan.so.6+0xb4917)
    #1 0x2b11e45064b6 in cpuinfo_x86_linux_init (/path/to/libonnxruntime.so.1.14.1+0x8684b6)

Direct leak of 576 byte(s) in 1 object(s) allocated from:
    #0 0x2b11dd429917 in __interceptor_calloc (/path/to/libasan.so.6+0xb4917)
    #1 0x2b11e4506496 in cpuinfo_x86_linux_init (/path/to/libonnxruntime.so.1.14.1+0x868496)

SUMMARY: AddressSanitizer: 1152 byte(s) leaked in 2 allocation(s).
```

ONNX Runtime doesn't call `cpuinfo_deinitialize`, but on the other hand that wouldn't help since [the implementation is empty](https://github.com/pytorch/cpuinfo/blob/f44a9dabb2192ffb203ddd0c71f6373c7d82faed/src/init.c#L60-L61).

It would be nice if `cpuinfo_deinitialize` could be implemented so that ONNX Runtime and other users can call it to deallocate the resources.

prashanthswami:
Users of the library can choose to release all resources the library may have allocated as part of a previous cpuinfo_initialize call. If no resources were previously allocated, this function does nothing.

Callers cannot re-initialize after the cpuinfo_deintialize call, as the library assumes initialization is a one-time operation per-process. Further, de-initializing can invalidate pointers previously acquired from cpuinfo (e.g. structures returned from cpuinfo_get_current_processor).

As such, callers should only call cpuinfo_deinitialize when:
1. There are no further uses of references returned from cpuinfo.
2. There are no further calls to cpuinfo_* functions.

Fixes #150

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -4,6 +4,9 @@
 	#include <pthread.h>
 #endif
 
+#include <inttypes.h>
+#include <string.h>
+
 #include <cpuinfo.h>
 #include <cpuinfo/internal-api.h>
 #include <cpuinfo/log.h>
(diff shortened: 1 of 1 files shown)


prashanthswami:
Review: @malfet

malfet (changes requested):
Hmm, there are not test to validate that it still works, is there?

Also, I see few `pthread_once` calls in `cpuinfo_initilize`, can you please explain how those resources would be freed?

Issue #19, opened 2019-01-05
[feature request] specify log-level during cpuinfo_initialize()


cdluminate:
Quoted from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=918177 :

```
libcpuinfo library selects log level during compile time.
1. It would be better if one could specify log-level during cpuinfo_initialize().
2. Anyway current compile-time setting is very annoing: on many processors it prints

Note (cpuinfo): Line power management: in /proc/cpuinfo is ignored: value contains only spaces
```

PR #209, opened 2023-11-29
Adjust log levels of /proc/cpuinfo parsing

prashanthswami:
There are a few steps in our parsing logic where we skip lines that don't match the expectations of the /proc/cpuinfo node. Reduce the log level of these lines to 'debug', as these are not generally errors and are noisy on systems that have unique cpuinfo key-value pairs.

When parsing logic encounters a higher-than-expected processor number, increase the level to warning, to indicate that an error may have occurred in the parsing step.

This does not fully address #19 but resolves the underlying noise reported.

### DIFF
--- a/src/arm/linux/cpuinfo.c
+++ b/src/arm/linux/cpuinfo.c
@@ -742,7 +742,7 @@ static bool parse_line(
 	}
 	/* Skip line if no ':' separator was found. */
 	if (separator == line_end) {
-		cpuinfo_log_info(
+		cpuinfo_log_debug(
 			"Line %.*s in /proc/cpuinfo is ignored: key/value separator ':' not found",
 			(int)(line_end - line_start),
 			line_start);
@@ -758,7 +758,7 @@ static bool parse_line(
 	}
 	/* Skip line if key contains nothing but spaces. */
 	if (key_end == line_start) {
-		cpuinfo_log_info(
+		cpuinfo_log_debug(
 			"Line %.*s in /proc/cpuinfo is ignored: key contains only spaces",
 			(int)(line_end - line_start),
 			line_start);
@@ -774,7 +774,7 @@ static bool parse_line(
 	}
 	/* Value part contains nothing but spaces. Skip line. */
 	if (value_start == line_end) {
-		cpuinfo_log_info(
+		cpuinfo_log_debug(
 			"Line %.*s in /proc/cpuinfo is ignored: value contains only spaces",
 			(int)(line_end - line_start),
 			line_start);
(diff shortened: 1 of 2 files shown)


prashanthswami:
Review: @malfet

fbarchard (approved):
thanks... these messages are not very interesting, so reducing it is a good idea

src/arm/linux/cpuinfo.c
```diff
@@ -802,7 +802,7 @@ static bool parse_line(
 			} else if (memcmp(line_start, "Revision", key_length) == 0) {
 				size_t value_length = value_end - value_start;
 				if (value_length > CPUINFO_REVISION_VALUE_MAX) {
-					cpuinfo_log_info(
+					cpuinfo_log_debug(
```
malfet, line 805: Hmm, this should be warning, isn't it? And if limit have become larger, should we just increase it instead?

src/arm/linux/cpuinfo.c
```diff
@@ -790,7 +790,7 @@ static bool parse_line(
 			} else if (memcmp(line_start, "Hardware", key_length) == 0) {
 				size_t value_length = value_end - value_start;
 				if (value_length > CPUINFO_HARDWARE_VALUE_MAX) {
-					cpuinfo_log_info(
+					cpuinfo_log_debug(
```
malfet, line 793: Same as below, imo this should stay an info or even be elevated to warning.

src/arm/linux/cpuinfo.c
prashanthswami: Upgraded to warning, the limit being fixed seems to be an artifact of the struct holding a fixed length array. We could change it to dynamically allocate? In any case, happy to do that in an isolated PR so that this one is focused on fixing the log levels.

prashanthswami: So I can't quite tell what the hardware revision or name is used for, because it's just an arbitrary string - surely, no one should be reading this value to make a decision? They should be requesting the micro-arch enums if they needed this information and needed to switch behavior on it.

Assuming that's true, then I'd suggest this should be debug - it's unfortunate that it cuts the name off, but shouldn't affect proper usage of the library / isn't something a user needs to actively fix.

malfet: Do you mind splitting this PR into two: let's downgrade others (as indeed empty lines or lines with no separators feels benign) and keep discussion on those in a separate issue PR.
 IMO if identifier needs to be truncated, it warrants a warning, but I've never seen it in practice

src/arm/linux/cpuinfo.c
prashanthswami: Splitting into separate PR as discussed above.

src/arm/linux/cpuinfo.c
prashanthswami: Ack, reverting these back to 'info' and creating separate PR.

malfet (approved):
Thank you for the updates, LGTM

PR #210, opened 2023-11-30
Fix chipset enum name to include 'vendor_'

prashanthswami:
The original change that introduced this should have used a consistent prefix for all enum types, for consistency sake.


malfet:
@prashanthswami can you please stop pinging me for reviews. This project at the moment needs a maintainer and I'm not sure I have a proper bandwith to be one (if I one, I should not have let original riscV change in without proper testing on my end, but I have not setup my riscv board yet)

prashanthswami:
Haven't gotten an invite yet, but did submit the form (and got confirmation the form was submitted), would you be able to help loop me in there? I'll ask around with some folks who've worked on this library internally to see if they can help provide support here.

PR #213, opened 2024-01-08
Run Bazel build in Github Actions

prashanthswami:
As some clients rely on the Bazel build, add a workflow to verify at least one Bazel target (linux-x86). Also, perform some minor cleanup to comments and target branches in our workflow files.


prashanthswami:
This should also address capturing issues like the ones called out in #213 202

scripts/local-bazel-build.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) Facebook, Inc. and its affiliates.
+# All rights reserved.
```
malfet, line 4: Feel free to remove this one and or replace it with your employer in the followup PR

malfet (approved):

scripts/local-bazel-build.sh
prashanthswami: Fixed!

PR #215, opened 2024-01-11
Fix RISC-V Linux build again

markdryan:
PR https://github.com/pytorch/cpuinfo/pull/204 broke the RISC-V build by including for a second time a header file that currently only exists in the RISC-V Android NDK.  The header is not yet available in mainstream Linux distributions.  The header in question, <sys/hwprobe.h>, is already included when building for Android at the top of riscv-hw.c so the second include is unnecessary and can be safely removed.

### DIFF
--- a/src/riscv/linux/riscv-hw.c
+++ b/src/riscv/linux/riscv-hw.c
@@ -14,7 +14,6 @@
 #endif
 
 #include <sched.h>
-#include <sys/hwprobe.h>
 
 #include <cpuinfo/log.h>
 #include <riscv/api.h>


markdryan:
@prashanthswami,  @malfet PR #204 broke the RISC-V build on mainline Linux distributions again.  It looks like a small error  was introduced resolving a conflict in riscv-hw.c.

prashanthswami:
Ouch - thank you. I need to go back and figure out why I missed this when running local tests. I should probably also introduce a builder that actually validates this consistently.

markdryan:
> Ouch - thank you. I need to go back and figure out why I missed this when running local tests. I should probably also introduce a builder that actually validates this consistently.

@prashanthswami I've had a go at adding some CI for riscv64 builds on Ubuntu 22.04

https://github.com/pytorch/cpuinfo/pull/219

Obviously, #219 isn't going to pass CI until this patch is merged and #219 is rebased.

prashanthswami (approved):

PR #218, opened 2024-01-12
ci: Add an Ubuntu:22.04 builder for RISC-V

markdryan:
cpuinfo is built for riscv64 using a riscv64 container.  binfmt_misc allows the riscv64 binaries in the container to be executed with QEMU. This is slower than cross compiling but as there's not that much code the build times are acceptable.  It should be easy to expand the matrix to add CI for other architectures not natively supported by github actions.


markdryan:
Sorry, pushed this too early.  Will resubmit once I've tested locally.

PR #219, opened 2024-01-12
ci: Add an Ubuntu:22.04 builder for RISC-V

markdryan:
cpuinfo is built for riscv64 using a riscv64 container.  binfmt_misc allows the riscv64 binaries in the container to be executed with QEMU. This is slower than cross compiling but as there's not that much code the build times are acceptable.  It takes just under 6 minutes for the full riscv64 github action to run.  We also have the option of running some of the built RISC-V binaries, e.g., unit tests, in the CI. It should be easy to expand the matrix to add CI for other architectures not natively supported by github actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,3 +93,26 @@ jobs:
         working-directory: ${{ github.workspace }}
         env:
           ANDROID_NDK: ${{ steps.setup-ndk.outputs.ndk-path }}
+  cmake-linux-qemu:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 40
+    strategy:
+      matrix:
+        build_props:
+          - [
+              "cmake-linux-riscv64",
+              "riscv64/ubuntu:22.04"
+          ]
+
+    name: ${{ matrix.build_props[0] }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@v3.0.0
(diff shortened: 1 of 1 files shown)


markdryan:
This PR will fail to build as the riscv64 builds are currently broken.  Once https://github.com/pytorch/cpuinfo/pull/215 is merged I'll rebase this patch and the CI should pass.

.github/workflows/build.yml
```diff
@@ -93,3 +93,27 @@ jobs:
...
+          - [
+              "cmake-linux-riscv64",
+              "riscv64/ubuntu:22.04"
+          ]
+
+    name: ${{ matrix.build_props[0] }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install qemu and setup binfmt_misc
+        run: |
```
luhenry, line 111: You should use the following instead:
```
- name: Setup QEMU
  uses: docker/setup-qemu-action@v3.0.0
  with:
    platforms: ${{ matrix.binfmt }} // binfmt is `riscv64` for RISC-V
```

markdryan: Thanks.  I'll give this a go.

markdryan: PR updated.

markdryan:
@prashanthswami I've rebased this PR against main, pulling in the fix to the riscv64 build.  The CI added for risc64 in this PR should pass now.

malfet (approved):
Thank you for adding the CI

PR #223, opened 2024-02-02
Correcting printing information errors

yichenchenyi:


tools/isa-info.c
```diff
@@ -13,7 +13,7 @@ int main(int argc, char** argv) {
 
 	printf("Scalar instructions:\n");
 #if CPUINFO_ARCH_X86
-	printf("\tx87 FPU: %s\n", cpuinfo_has_x86_fpu() ? "yes" : "no");
+	printf("\tx86 FPU: %s\n", cpuinfo_has_x86_fpu() ? "yes" : "no");
```
xuhancn, line 16: FPU is x87, here is right: https://linasm.sourceforge.net/docs/instructions/fpu.php

yichenchenyi: > FPU is x87, here is right: https://linasm.sourceforge.net/docs/instructions/fpu.php

thanks I made a mistake

PR #225, opened 2024-02-24
cachebreak

fbarchard:
Add a missing break to cpuinfo_x86_decode_cache_descriptor

PR #226, opened 2024-02-26
cpuinfo_x86_normalize_brand_string

fbarchard:
Fix cpuinfo_x86_normalize_brand_string unannotated fall-through warning

### DIFF
--- a/src/x86/name.c
+++ b/src/x86/name.c
@@ -612,6 +612,8 @@ uint32_t cpuinfo_x86_normalize_brand_string(const char raw_name[48], char normal
 					break;
 				case '@':
 					parser_state.frequency_separator = char_ptr;
+					*char_ptr = ' ';
+					break;
 				case '\0':
 				case '\t':
 					*char_ptr = ' ';


malfet (changes requested):
Can you perhaps provide an example of the config when it raises this warning, as cpuinfo code is explicitly marked as C99 compliant, so CMake selects `-Wno-implicit-falltrhough` flag to suppress this warning.

But if you are using C23 compiler, please modify this PR to use [`[[fallthrough]]`](https://en.cppreference.com/w/c/language/attributes/fallthrough) attribute (guarded with proper ifdefs) to mark this falltrhough as expected

fbarchard:
The issue was reported as XNNPack not building with the -Wimplicit-fallthrough flag 
https://github.com/google/XNNPACK/issues/6047
But when I apply the flag in XNNPack build the pytorch/cpuinfo has 2 build warnings.
The first was a legitimate bug and fixed in https://github.com/pytorch/cpuinfo/pull/225
This one is not a bug, but the other cases are handled the same was as this PR.

XNNPack requires C11 but can work on VS2012 or GCC 4.9.
If this PR is taken, x86 builds cleanly, but for ARM are more, but I didnt spot legitimate bugs, so I wasnt going to do the fallthru directives.

PR #227, opened 2024-03-04
Update cpuinfo.h: add support for ARM64EC

snnn:
To fix a build error:
```
src\arm\windows\init-by-logical-sys-info.c(309,17): Error C2065: 'cpuinfo_uarchs': undeclared identifier
```
In short, Arm64EC is a new application binary interface (ABI) for apps running on Arm devices with an x64 like ABI.

Reference: 
1. [VC++ predefined-macros](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170)
2. [ARM64EC overview](https://learn.microsoft.com/en-us/windows/arm/arm64ec)

### DIFF
--- a/include/cpuinfo.h
+++ b/include/cpuinfo.h
@@ -18,7 +18,7 @@
 #define CPUINFO_ARCH_X86 1
 #endif
 
-#if defined(__x86_64__) || defined(__x86_64) || defined(_M_X64) || defined(_M_AMD64)
+#if defined(__x86_64__) || defined(__x86_64) || (defined(_M_X64) && !defined(_M_ARM64EC)) || (defined(_M_AMD64) && !defined(_M_ARM64EC))
 #define CPUINFO_ARCH_X86_64 1
 #endif
 
@@ -26,7 +26,7 @@
 #define CPUINFO_ARCH_ARM 1
 #endif
 
-#if defined(__aarch64__) || defined(_M_ARM64)
+#if defined(__aarch64__) || defined(_M_ARM64) || defined(_M_ARM64EC)
 #define CPUINFO_ARCH_ARM64 1
 #endif
 


include/cpuinfo.h
```diff
@@ -18,15 +18,15 @@
 #define CPUINFO_ARCH_X86 1
 #endif
 
-#if defined(__x86_64__) || defined(__x86_64) || defined(_M_X64) || defined(_M_AMD64)
+#if defined(__x86_64__) || defined(__x86_64) || (defined(_M_X64) && !defined(_M_ARM64EC)) || (defined(_M_AMD64) && !defined(_M_ARM64EC))
```
xuhancn, line 21: I checked https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170
MSVC architecture related defination is complex, suggest to add more comments.

xuhancn (approved):

include/cpuinfo.h
malfet, line 21: Defining `CPUINFO_ARCH_X86_64` for `_M_ARM64EC` feels wrong, please explain why this is needed

malfet (changes requested):
It feels wrong to define CPUINFO_ARCH_X86_64 on ARM platform, please explain why this is necessary for Win ARM64

include/cpuinfo.h
snnn: _M_X64 Defined as the integer literal value 100 for compilations that target x64 processors or ARM64EC. So _M_X64  could be 1 when the CPU is actually an ARM CPU.  
Here the condition is `defined(_M_X64) && !defined(_M_ARM64EC))` . When _M_ARM64EC is 1, the expression is false.

snnn:
@malfet , would you mind taking another look at this PR? We need this change to make https://github.com/microsoft/onnxruntime work for ARM64EC builds. A lot of Microsoft applications ship with ARM64EC ABI.

include/cpuinfo.h
snnn: > It feels wrong to define CPUINFO_ARCH_X86_64 on ARM platform, 

It is the problem I am facing and fixing. _M_X64 is defined when the ABI is ARM64EC. So, without this change, CPUINFO_ARCH_X86_64  is defined when the target ABI is  ARM64EC which is target for ARM64 CPUs.

PR #172, opened 2023-08-07
add FreeBSD support

cyyever:

PR #230, opened 2024-03-21
Up-to-date FreeBSD support

yurivict:
Same as https://github.com/pytorch/cpuinfo/pull/172, updated to the latest cpuinfo revision.


fbarchard (commented):
I'm not an expert in the OS layer.  The code looks okay.  Is it necessary to have freebsd instead of linux?

yurivict:
@Maratyszcza 

Hi Marat,

This is an updated pull request that works fine ion FreeBSD.
The FreeBSD port replaces cpuinfo with this revision and PyTorch works fine.

Could you please merge it?

Thanks,
Yuri

martymac:
Hello,

Any news on that PR ? FYI, I used it to update our pcsx2 port and it works fine here.

Cheers,

Ganael.

src/x86/freebsd/init.c
```diff
@@ -0,0 +1,402 @@
...
+	if (clusters == NULL) {
+		cpuinfo_log_error(
+		    "failed to allocate %zu bytes for descriptions of %" PRIu32
+		    " core clusters",
+		    freebsd_topology.packages * sizeof(struct cpuinfo_cluster),
+		    freebsd_topology.packages);
+		goto cleanup;
+	}
+	packages =
+	    calloc(freebsd_topology.packages, sizeof(struct cpuinfo_package));
```
fbarchard, line 66: Is packages the reason you're wanting a branch from linux?
The latest linux kernel on android and riscv requires packages support so it would be best to use common code to avoid freebsd diverging in the future.

PR #240, opened 2024-04-23
Use cluster_cpus_list to detect clusters for ARM

prashanthswami:
On newer kernels (6.x+) a separate topology node exists to express the list of cpus associated with each cluster. The cluster topology node explicitly refers to cores that share some resources (e.g. cache) as opposed to 'core_siblings', which only indicate they are on the same physical package.

This distinction is relevant on 6.x, where some devices may have all their cores on a single physical package but cores do not all share the same resources on that package. In these cases, the existing logic incorrectly reports that there is a single cluster (because all cores are siblings), even though not all cores share the same resources.

Test: Verified on ARM64 Android device that has multiple clusters, where core_siblings reports all cores are siblings.
Test: Added mock hardware tests, verified pass on ARM64 hardware.

### DIFF
--- a/test/mock/pixel-8.h
+++ b/test/mock/pixel-8.h
@@ -1,0 +1,1294 @@
+struct cpuinfo_mock_file filesystem[] = {
+	{.path = "/proc/cpuinfo",
+	 .size = 3717,
+	 .content =
+		 "processor\t: 0\n"
+		 "BogoMIPS\t: 49.15\n"
+		 "Features\t: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 sve asimdfhm dit uscat ilrcpc flagm ssbs sb paca pacg dcpodp sve2 sveaes svepmull svebitperm svesha3 svesm4 flagm2 frint svei8mm svebf16 i8mm bti\n"
+		 "CPU implementer\t: 0x41\n"
+		 "CPU architecture: 8\n"
+		 "CPU variant\t: 0x1\n"
+		 "CPU part\t: 0xd46\n"
+		 "CPU revision\t: 1\n"
+		 "\n"
+		 "processor\t: 1\n"
+		 "BogoMIPS\t: 49.15\n"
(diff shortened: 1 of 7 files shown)


fbarchard (approved):

digantdesai:
cc @kimishpatel FYI

digantdesai:
LGTM, can you include some kernel doc or commit which changes the behavior in the PR description or in the comments here for future reference? Thanks.

test/mock/pixel-8.h
```diff
@@ -0,0 +1,1294 @@
...
+		    "1098000 4202\n"
+		    "1197000 2559\n"
+		    "1328000 4948\n"
+		    "1425000 1421\n"
+		    "1548000 1288\n"
+		    "1704000 9901\n"},
+	{.path = "/sys/devices/system/cpu/cpu0/cpufreq/stats/total_trans", .size = 7, .content = "115798\n"},
+	{.path = "/sys/devices/system/cpu/cpu0/topology/core_id", .size = 2, .content = "0\n"},
+	{.path = "/sys/devices/system/cpu/cpu0/topology/core_siblings_list", .size = 2, .content = "0\n"},
+	{.path = "/sys/devices/system/cpu/cpu0/topology/cluster_cpus", .size = 4, .content = "00f\n"},
```
digantdesai, line 127: I guess this replaces `core_siblings` on newer kernels?

prashanthswami:
For posterity, here's [the commit that introduced cluster_cpus](https://github.com/torvalds/linux/commit/c5e22feffdd736cb02b98b0f5b375c8ebc858dd4). I should stress that this didn't really "replace" core_sibilings_list, so much as it "created a more specific node that identifies a true 'cluster'". We were using core_siblings_list before as a proxy, but if you take a look at the [implementation](https://github.com/torvalds/linux/blob/c5e22feffdd736cb02b98b0f5b375c8ebc858dd4/drivers/base/arch_topology.c#L631) you'll see that it's simply checking to see if they are on the same package and then marks them as a core sibling. There's no *explicit* guarantee that they share an internal system bus (though one might imagine that there are probably many examples of SoCs that have N-cores in a single cluster, on a single package).

That is why that we also retain the fallback path - if we cannot use the ACPI-defined clusters, at worst we'll still mirror the behavior we've always had and just assume we're on a chipset where the cores are in a single cluster as before.

digantdesai:
Thank you @prashanthswami, very helpful.

PR #244, opened 2024-05-22
Fix build error for MinGW aarch64

hans00:
MinGW's aarch64 header define is no `Cache.GroupCount`


malfet (approved):
Looks good to me, but please fix clang-format errors

PR #255, opened 2024-08-06
Add a API to check SVE Length support on ARM CPU.

maajidkhann:
This pull request introduces a new feature to the cpuinfo library that adds an API to return the maximum supported Scalable Vector Extension (SVE) vector length on the given ARM CPU. This enhancement will allow users to query and determine the maximum SVE vector lengths on a given ARM CPU, providing better insights and flexibility for optimizing applications that utilize SVE.

**Key Features:**
**New API Function:**

Introduces a single API function - cpuinfo_get_max_arm_sve_length() that returns the maximum SVE Vector Length supported on the given ARM CPU.

The function is designed to be easy to integrate with existing code in other projects like PyTorch (https://github.com/pytorch/pytorch/pull/119571/) and provides a straightforward interface for querying SVE VL.

**Here's the sample output on SVE supported instance:**

**Query:**
test_cpp_aug7

**Output on SVE256 supported Hardware - Graviton3:**
test_cpp_output

### DIFF
--- a/include/cpuinfo.h
+++ b/include/cpuinfo.h
@@ -1670,6 +1670,7 @@ struct cpuinfo_arm_isa {
 	bool sve;
 	bool sve2;
 	bool i8mm;
+	int svelen;
 #endif
 	bool rdm;
 	bool fp16arith;
@@ -2039,6 +2040,30 @@ static inline bool cpuinfo_has_arm_sve2(void) {
 	return cpuinfo_isa.sve2;
 #else
 	return false;
+#endif
+}
+
+static inline bool cpuinfo_support_arm_sve128(void) {
+#if CPUINFO_ARCH_ARM || CPUINFO_ARCH_ARM64
+    return (cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen >= 16);
+#else
+    return false;
+#endif
+}
+
+static inline bool cpuinfo_support_arm_sve256(void) {
+#if CPUINFO_ARCH_ARM || CPUINFO_ARCH_ARM64
+    return (cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen >= 32);
+#else
+    return false;
+#endif
+}
+
+static inline bool cpuinfo_support_arm_sve512(void) {
+#if CPUINFO_ARCH_ARM || CPUINFO_ARCH_ARM64
+    return (cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen >= 64);
+#else
+    return false;
 #endif
 }
 
--- a/include/cpuinfo.h
+++ b/include/cpuinfo.h
@@ -1670,7 +1670,7 @@ struct cpuinfo_arm_isa {
 	bool sve;
 	bool sve2;
 	bool i8mm;
-	unsigned int svelen;
+	uint32_t svelen;
 #endif
 	bool rdm;
 	bool fp16arith;
@@ -2044,14 +2044,11 @@ static inline bool cpuinfo_has_arm_sve2(void) {
 }
 
 // Function to get the max SVE vector length on ARM CPU's which support SVE.
-static inline int cpuinfo_get_max_arm_sve_length(void) {
+static inline uint32_t cpuinfo_get_max_arm_sve_length(void) {
 #if CPUINFO_ARCH_ARM64
-	if ((cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen > 0))
-		return cpuinfo_isa.svelen;
-	else
-		return -1; // Returns -1 on Non SVE supported ARM CPU's.
+	return cpuinfo_isa.svelen * 8; // bytes * 8 = bit length(vector length)
 #else
-	return -1; // Returns -1 on Non ARM CPU's.
+	return 0;
 #endif
 }
 
--- a/include/cpuinfo.h
+++ b/include/cpuinfo.h
@@ -1670,7 +1670,7 @@ struct cpuinfo_arm_isa {
 	bool sve;
 	bool sve2;
 	bool i8mm;
-	int svelen;
+	unsigned int svelen;
 #endif
 	bool rdm;
 	bool fp16arith;
@@ -2043,27 +2043,15 @@ static inline bool cpuinfo_has_arm_sve2(void) {
 #endif
 }
 
-static inline bool cpuinfo_support_arm_sve128(void) {
-#if CPUINFO_ARCH_ARM || CPUINFO_ARCH_ARM64
-    return (cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen >= 16);
+// Function to get the max SVE vector length on ARM CPU's which support SVE.
+static inline int cpuinfo_get_max_arm_sve_length(void) {
+#if CPUINFO_ARCH_ARM64
+	if ((cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen > 0))
+		return cpuinfo_isa.svelen;
+	else
+		return -1; // Returns -1 on Non SVE supported ARM CPU's.
 #else
-    return false;
-#endif
-}
-
-static inline bool cpuinfo_support_arm_sve256(void) {
-#if CPUINFO_ARCH_ARM || CPUINFO_ARCH_ARM64
-    return (cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen >= 32);
-#else
-    return false;
-#endif
-}
-
-static inline bool cpuinfo_support_arm_sve512(void) {
-#if CPUINFO_ARCH_ARM || CPUINFO_ARCH_ARM64
-    return (cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen >= 64);
-#else
-    return false;
+	return -1; // Returns -1 on Non ARM CPU's.
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


maajidkhann:
@malfet @digantdesai  Please review this PR. This will make it easy for people to use API's from CPUINFO into PyTorch to get SVE vector length info from the hardware.

src/arm/linux/aarch64-isa.c
```diff
@@ -151,4 +156,15 @@ void cpuinfo_arm64_linux_decode_isa_from_proc_cpuinfo(
 	if (features & CPUINFO_ARM_LINUX_FEATURE_ASIMDFHM) {
 		isa->fhm = true;
 	}
+
+    #define PR_SVE_GET_VL 51
+    #define PR_SVE_VL_LEN_MASK 0xffff
```
malfet, line 161: Don't you want to add something like `#ifndef PR_SVE_GET_VL` guard here? And also, please mention which system headers they should be defined in

tools/isa-info.c
```diff
@@ -172,6 +172,11 @@ int main(int argc, char** argv) {
 	printf("\tARM SVE: %s\n", cpuinfo_has_arm_sve() ? "yes" : "no");
 	printf("\tARM SVE 2: %s\n", cpuinfo_has_arm_sve2() ? "yes" : "no");
 
+	printf("ARM SVE Capabilities:\n");
+	printf("\tSVE 128-bit: %s\n", cpuinfo_support_arm_sve128() ? "Yes" : "No");
+	printf("\tSVE 256-bit: %s\n", cpuinfo_support_arm_sve256() ? "Yes" : "No");
+	printf("\tSVE 512-bit: %s\n", cpuinfo_support_arm_sve512() ? "Yes" : "No");
```
malfet, line 178: https://developer.arm.com/documentation/101458/2404/Compiler-options/-msve-vector-bits- seems to indicate, that SVE length can be up to 2048? If this is the case, wouldn't it be better to add `cpuinfo_get_arm_sve_lengh()` method rather than enum? Also, if it's always a power of two, wouldn't it be easier if it returned a bitshifts?

include/cpuinfo.h
```diff
@@ -1670,6 +1670,7 @@ struct cpuinfo_arm_isa {
 	bool sve;
 	bool sve2;
 	bool i8mm;
+	int svelen;
```
malfet, line 1673: Why signed type? Is `-1` used to indicate `unsupported`?

malfet (changes requested):
Please fix formatting (see lint failure in CI)
Also, all links to the images seems to be broken
And last but not least, please explain why boolean `has_sve128` are better than single `get_max_sve_len` API?

src/arm/linux/aarch64-isa.c
```diff
@@ -151,4 +156,15 @@ void cpuinfo_arm64_linux_decode_isa_from_proc_cpuinfo(
...
+    int ret = prctl(PR_SVE_GET_VL);
+    if (ret < 0) {
+        perror("prctl(PR_SVE_GET_VL) failed");
```
malfet, line 164:
```suggestion
        cpuinfo_log_error("prctl(PR_SVE_GET_VL) failed");
```

src/arm/linux/aarch64-isa.c
maajidkhann: Done.

tools/isa-info.c
maajidkhann: Agree. Done.
My intention in the first place was to introduce functions like how AVX has done for AVX2 and AVX512 which returns a bool and for SVE as only VL=128, 256 and 512 are relevant in ARM ecosystem, I added three different functions.

But it makes sense to just add a single function that returns maximum VL supported on that hardware and now the user can work out the logic required with the value returned.

include/cpuinfo.h
maajidkhann: As that was the initial thought, -1 = unsupported. Now I have changed it to unsigned int as we don't deal with negative values and svelen value always remains >=0 in our context.

src/arm/linux/aarch64-isa.c
maajidkhann: Done

maajidkhann:
> Please fix formatting (see lint failure in CI) Also, all links to the images seems to be broken And last but not least, please explain why boolean `has_sve128` are better than single `get_max_sve_len` API?

Addresses all the comments. Lint issues also have been now fixed.

include/cpuinfo.h
```diff
@@ -2042,6 +2043,18 @@ static inline bool cpuinfo_has_arm_sve2(void) {
 #endif
 }
 
+// Function to get the max SVE vector length on ARM CPU's which support SVE.
+static inline int cpuinfo_get_max_arm_sve_length(void) {
+#if CPUINFO_ARCH_ARM64
+	if ((cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen > 0))
+		return cpuinfo_isa.svelen;
```
malfet, line 2050: If I understood your previous diff correctly, svelen is returned in bytes, but API developers usually think about it in bitlength, therefore it must be multiplied by 8, shouldn't it?
```suggestion
		return cpuinfo_isa.svelen * 8;
```

include/cpuinfo.h
```diff
@@ -1670,6 +1670,7 @@ struct cpuinfo_arm_isa {
 	bool sve;
 	bool sve2;
 	bool i8mm;
+	unsigned int svelen;
```
malfet, line 1673:
```suggestion
	uint32_t svelen;
```

include/cpuinfo.h
```diff
@@ -2042,6 +2043,18 @@ static inline bool cpuinfo_has_arm_sve2(void) {
...
+static inline int cpuinfo_get_max_arm_sve_length(void) {
+#if CPUINFO_ARCH_ARM64
+	if ((cpuinfo_isa.sve || cpuinfo_isa.sve2) && (cpuinfo_isa.svelen > 0))
+		return cpuinfo_isa.svelen;
+	else
+		return -1; // Returns -1 on Non SVE supported ARM CPU's.
+#else
+	return -1; // Returns -1 on Non ARM CPU's.
+#endif
+}
```
malfet, line 2056: Why not keep it simple? One does not need a check, on whether or not svelen is supported, as it should be initialized to 0, shouldn't it?
```suggestion
static inline uint32_t cpuinfo_get_max_arm_sve_length(void) {
#if CPUINFO_ARCH_ARM64
	return cpuinfo_isa.svelen;
#else
	return 0;
#endif
}
```

src/arm/linux/aarch64-isa.c
```diff
@@ -151,4 +151,25 @@ void cpuinfo_arm64_linux_decode_isa_from_proc_cpuinfo(
...
+
+	int ret = prctl(PR_SVE_GET_VL);
+	if (ret < 0) {
+		cpuinfo_log_error("prctl(PR_SVE_GET_VL) failed");
+		isa->svelen = 0; // Assume no SVE support if the call fails
+	} else {
+		// Mask out the SVE vector length bits
+		isa->svelen = ret & PR_SVE_VL_LEN_MASK;
+	}
+#endif
```
malfet, line 174:
```suggestion
#else
        isa->svelen = 0;
#endif
```

src/arm/linux/aarch64-isa.c
```diff
@@ -151,4 +151,25 @@ void cpuinfo_arm64_linux_decode_isa_from_proc_cpuinfo(
 	if (features & CPUINFO_ARM_LINUX_FEATURE_ASIMDFHM) {
 		isa->fhm = true;
 	}
+
+#if defined(__linux__)
+#include <sys/prctl.h>
```
malfet, line 156: Adding `#include` directive in the body of the function feels very fragile, can you please move it to the top level of the file?

src/arm/linux/aarch64-isa.c
```diff
@@ -151,4 +151,25 @@ void cpuinfo_arm64_linux_decode_isa_from_proc_cpuinfo(
 	if (features & CPUINFO_ARM_LINUX_FEATURE_ASIMDFHM) {
 		isa->fhm = true;
 	}
+
+#if defined(__linux__)
```
malfet, line 155: Why does it needs `#ifdef __linux__`, as this file should only be compiled for Linux anyway, isn't it?

malfet (approved):
Thank you for all the updates, looks good, but please consider addressing few minor remarks

include/cpuinfo.h
maajidkhann: Yes its better to return bit length directly compared to bytes to avoid confusion to the end users.
Changes done.

include/cpuinfo.h
maajidkhann: Done

include/cpuinfo.h
maajidkhann: Agree, changes done.

src/arm/linux/aarch64-isa.c
maajidkhann: This change is not required anymore as we have removed the following check - **#if defined(__linux__)** (Not required as the code is compiled for Linux anyways).

src/arm/linux/aarch64-isa.c
maajidkhann: Done

src/arm/linux/aarch64-isa.c
maajidkhann: Agree. Removed this macro.

maajidkhann:
> Thank you for all the updates, looks good, but please consider addressing few minor remarks

All changes made as suggested.

Here's the test output, it now returns 256 on Graviton 3 EC2 instance.

test_output_cpuifo_aug7

maajidkhann:
I was also able to call this new API from Pytorch source code and access the vector length correctly.

image

Issue #206, opened 2023-11-22
Add a Github workflow to build Android RISC-V, when an NDK is available


prashanthswami:
Similar to our other Android targets in the workflows, we should add an Android RISC-V build target. This isn't currently trivial, because the Android NDK w/RISC-V support is only in canary, so needs to be downloaded off the Android public builders, as opposed to the current workflow actions that expect a stable release.

We could hack in support for that, or alternatively just wait for a formal release to be made.

PR #256, opened 2024-08-19
Add android-riscv64 build to workflows

prashanthswami:
The Android NDK supports building for RISC-V as of version r27. Add it to the build workflows here to improve coverage of the RISC-V code.

Fixes: #206

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,7 +73,7 @@ jobs:
   cmake-android:
     strategy:
       matrix:
-        script: [android-arm64-build.sh, android-armv7-build.sh, android-x86-build.sh]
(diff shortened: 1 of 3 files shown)


malfet (approved):

rrwinterton:
There are public documents on AVX 10.1 and AVX 10.2. There ISA is available here: https://www.intel.com/content/www/us/en/content-details/836199/intel-advanced-vector-extensions-10-2-intel-avx10-2-architecture-specification.html which you can download AVX 10.1 and AVX 10.2 from the drop down version. Currently Granite Rapids has AVX 10.1 and near term future CPU will have both AVX 10.1 and AVX 10.2 "converged SIMD ISA. Per Franks request I will add the AVX 10.2 with several new instructions and simplifications.

PR #257, opened 2024-08-27
Add detection for `sme` on `aarch64`.

gonnet:
This is based on the ARM ComputeLibrary implementation [here](https://github.com/ARM-software/ComputeLibrary/blob/f1929dc994d8e5afae5c77ca66446344119a8592/src/common/cpuinfo/CpuIsaInfo.cpp#L75).

Not quite sure how to implement tests for this though.


digantdesai:
> Not quite sure how to implement tests for this though.

Can we do it through CPUINFO_MOCK?

fbarchard (approved):
consider adding apple support (for M4)

Typically with new registers, linux / Windows kernel support is required?  Do we need a check for OS support?

malfet (approved):
Looks good to me, though I wonder if adding more and more of those flags makes sense vs allowing user to query those directly

PR #261, opened 2024-09-19
read cpu info from cpuid.info file.

bunnitha:
Instrumented cpuinfo to read the populated cpu info data from cpuid.info file during initialization.
Consumers of the library or OEMs can trigger a oneshot service - cpuinfo-svc - during device boot to link libcpuinfo and dump all cpu info related to processor, model, arch and isa capabilities into cpuid.info file. This arrangement will limit cpuid() intrinsic calls (which causes VM-Exit events) to only once during init of libcpuinfo at boot by cpuinfo-svc.

### DIFF
--- a/src/x86/api.h
+++ b/src/x86/api.h
@@ -33,6 +33,22 @@ struct cpuinfo_x86_caches {
 	uint32_t prefetch_size;
 };
 
+struct cpuinfo_x86_tlbs {
+	struct cpuinfo_tlb itlb_4KB;
+	struct cpuinfo_tlb itlb_2MB;
+	struct cpuinfo_tlb itlb_4MB;
+	struct cpuinfo_tlb dtlb0_4KB;
+	struct cpuinfo_tlb dtlb0_2MB;
+	struct cpuinfo_tlb dtlb0_4MB;
+	struct cpuinfo_tlb dtlb_4KB;
+	struct cpuinfo_tlb dtlb_2MB;
+	struct cpuinfo_tlb dtlb_4MB;
+	struct cpuinfo_tlb dtlb_1GB;
+	struct cpuinfo_tlb stlb2_4KB;
+	struct cpuinfo_tlb stlb2_2MB;
+	struct cpuinfo_tlb stlb2_1GB;
+};
+
 struct cpuinfo_x86_model_info {
 	uint32_t model;
 	uint32_t family;
@@ -61,21 +77,7 @@ struct cpuinfo_x86_processor {
 	int linux_id;
 #endif
 	struct cpuinfo_x86_caches cache;
-	struct {
-		struct cpuinfo_tlb itlb_4KB;
-		struct cpuinfo_tlb itlb_2MB;
-		struct cpuinfo_tlb itlb_4MB;
-		struct cpuinfo_tlb dtlb0_4KB;
-		struct cpuinfo_tlb dtlb0_2MB;
-		struct cpuinfo_tlb dtlb0_4MB;
-		struct cpuinfo_tlb dtlb_4KB;
-		struct cpuinfo_tlb dtlb_2MB;
-		struct cpuinfo_tlb dtlb_4MB;
-		struct cpuinfo_tlb dtlb_1GB;
-		struct cpuinfo_tlb stlb2_4KB;
-		struct cpuinfo_tlb stlb2_2MB;
-		struct cpuinfo_tlb stlb2_1GB;
-	} tlb;
+	struct cpuinfo_x86_tlbs tlb;
 	struct cpuinfo_x86_topology topology;
 	char brand_string[CPUINFO_PACKAGE_NAME_MAX];
 };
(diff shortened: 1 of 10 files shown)


bunnitha:
Hi, requesting concerned moderator to take a look at this proposal and review the patch.
Intel wants Pytorch to adopt this solution so that it can later land in Chrome ARCVM tree.

src/x86/init.c
```diff
@@ -7,11 +7,41 @@
...
+/*
+ * Structs cpuinfo_x86_isa, cpuinfo_x86_model_info and cpuinfo_x86_processor
+ * are dumped into cpuid.info file by cpuinfo-svc service on boot.
+ * Whenever libcpuinfo is initialized by any application, it will attempt to
+ * read cpuid.info file, if it exists, and the binary dump in the file is
+ * copied to these structs. If the file doesn't exist, we will fall back to
+ * original method of calling __cpuid() intrinsics to get the info.
+ * By this method, we can avoid __cpuid() calls during each init of cpuinfo.
+ * This will reduce VMX events in virtualized environments thereby
+ * improving performance.
```
digantdesai, line 21: Do you have some data? Please share it in the PR summary?

src/x86/init.c
```diff
@@ -7,11 +7,41 @@
...
+ */
+
+
+#ifdef __ANDROID__
```
digantdesai, line 25: can we create android specific dir like in arm rather than `#ifdef`?

src/x86/linux/cpuinfo.c
```diff
@@ -217,3 +217,126 @@ bool cpuinfo_x86_linux_parse_proc_cpuinfo(
 	return cpuinfo_linux_parse_multiline_file(
 		"/proc/cpuinfo", BUFFER_SIZE, (cpuinfo_line_callback)parse_line, &state);
 }
+
+#ifdef __ANDROID__
```
digantdesai, line 221: same here.

digantdesai (changes requested):
Thank you for your contribution.

(1) tlb refactor should be in a different commit
(2) please add some tests for cpuid features
(3) update PR summary for performance data and rationale for Android vm-exit overhead reduction.

bunnitha:
separated tlb refactor codes and cpu info caching.

src/x86/init.c
bunnitha: #ifdef __ANDROID__ has already been used in the project.
Ref: https://github.com/pytorch/cpuinfo/blob/aaac07ee499895770c89163ce0920ef8bb41ed23/src/x86/cache/descriptor.c#L1224

src/x86/init.c
bunnitha: Its little tricky to count the exact number of VM-Exits caused by CPUID calls.
For this effect, we have written a script to run an Android (dummy) service
that loads libcpuinfo - first in original version and next in modified version.
libcpuinfo makes around 4 __cpuid() calls on each initialization.
So, to amplify the effect, we have run the dummy service 1000 times and timed the total
execution (4000 __cpuid() calls and 8000 VMX events) in nano second precision.
This test shows that there is a reduction of approx. 0.6% in time to execute libcpuinfo.

Also, we attempted to record the KVM event statistics using perf tool.
Perf records the stats during the execution of same script.
The data shows that approx. 20,000 kvm_cpuid events reduced in modified libcpuinfo calls.
We cannot assume that all the CPUID calls are contributed by our script.\
We are only making a safe assumption that the proportional reduction is affected
by the modification we did. We had made all efforts to keep the test conditions
similar in both cases.

Please find attached the excel sheet containing the test data.
[libcpuinfo_enhancement_test_data.xlsx](https://github.com/user-attachments/files/18777275/libcpuinfo_enhancement_test_data.xlsx)

digantdesai (changes requested):
Hello, apologies for a delay, can you help me understand, 

(1) why does it have to be `__ANDROID__` specific. And Why can't a Linux user use cpuinfo-svc? 
(1.1) Does it have to be x86 specific? We don't have to use it for other ISAs, I don't have a too strong preference here.
(2) why a user facing "app" is in the `src/x86/linux/` vs `cpuinfo/tools`?

bunnitha:
Thank you @digantdesai for your review.

Answer to (1) and (2):
The solution can be generic.
We have had it devised for ARCVM (Virtualized Android) on CrOS.
The deployment and validation have only been done on Android in x86 target (others are out of our scope).
To make sure the rest of the world don't break, we applied the __ANDROID__ restriction.

digantdesai:
> (2) why a user facing "app" is in the src/x86/linux/ vs cpuinfo/tools?
I am ok if you restrict caching to android + x86, but please move the "app" out of the src dir.

bunnitha:
> > (2) why a user facing "app" is in the src/x86/linux/ vs cpuinfo/tools?
> > I am ok if you restrict caching to android + x86, but please move the "app" out of the src dir.

Agreed!
As suggested, app moved into tools/ folder

src/x86/api.h
```diff
@@ -109,40 +111,15 @@ CPUINFO_INTERNAL void cpuinfo_x86_detect_cache(
...
-	struct cpuinfo_tlb* dtlb0_2MB,
-	struct cpuinfo_tlb* dtlb0_4MB,
-	struct cpuinfo_tlb* dtlb_4KB,
-	struct cpuinfo_tlb* dtlb_2MB,
-	struct cpuinfo_tlb* dtlb_4MB,
-	struct cpuinfo_tlb* dtlb_1GB,
-	struct cpuinfo_tlb* stlb2_4KB,
-	struct cpuinfo_tlb* stlb2_2MB,
-	struct cpuinfo_tlb* stlb2_1GB,
-	uint32_t* prefetch_size);
```
digantdesai, line 145: did we drop this?

src/x86/linux/cpuinfo.c
```diff
@@ -217,3 +217,126 @@ bool cpuinfo_x86_linux_parse_proc_cpuinfo(
...
+    struct cpuinfo_x86_cpuid_info* x86_cpuid_info)
+{
+    cpuinfo_log_debug("reading cpuid.info file");
+    FILE* fd_info = fopen(CPUID_INFO_FILE, "rb");
+    if (!fd_info) {
+        cpuinfo_log_warning("failed to open file %s: %s", CPUID_INFO_FILE, strerror(errno));
+        return false;
+    }
+
+    int ret = fread(x86_cpuid_info, sizeof(struct cpuinfo_x86_cpuid_info), 1, fd_info);
```
digantdesai, line 232: Is it possible to read stale file where we may read a different version of the cpuid_info struct. Can we add a method to validate the struct after loading and before consuming?

digantdesai (commented):
I think we are getting pretty close. Left some minor comments. Thanks!

src/x86/api.h
bunnitha: No. I have moved them to a separate commit as you had suggested earlier [here](https://github.com/pytorch/cpuinfo/pull/261/commits/f6c059f962f86aa50a8217048a320f40d2ce3ec9)

Now, there are 2 commits in this PR. one for the tlb refactor and the other one the original solution to cache cpuid info.

src/x86/linux/cpuinfo.c
bunnitha: Thanks for the real good reviews. We appreciate your effort @digantdesai 
--------------------------------------------------------------------
Q. Is it possible to read stale file where we may read a different version of the cpuid_info struct.

Re: This is actually taken care of in Android environment.
The user facing app - cpuinfo-svc is triggered at every boot.
This will achieve 3 objectives:
1. Any change in hardware is taken care (both bare metal and virtualized environment). The cpuid data in the file will always be latest.
2. The file will always be rewritten making sure that there is no stale data or older version file exists.
3. Android file ownership makes sure that no external element can modify the content of the file at runtime.
----------------------------------------------------------------------
Q. Can we add a method to validate the struct after loading and before consuming?

Re: If you can think of any corner cases or scenario which we need to take care of, let us know; we can device some mechanism to validate the integrity of the data in the file.

src/x86/api.h
digantdesai: No I mean the `prefetch_size` variable.

src/x86/linux/cpuinfo.c
digantdesai: > The cpuid data in the file will always be latest.

This assumes things about set up outside of this library. I am thinking perhaps we can add a trivial check like, adding a new `size_t size` field at the end of the struct that is being stored/restored from a file. 

When we instantiate, we can do `struct.size = sizeof(struct)` and when create a new one from a file, we can assert `struct.size == sizeof(struct)`.

src/x86/api.h
bunnitha: *prefetch_size ( mapped to uint32_t* log2_package_cores_max ) is passed to the function to collect value to assign to processor->topology.core_bits_length.
image
for better readability, &topology is passed in place of *prefetch_size and got the value assigned directly.
image

digantdesai:
PR LGTM. 
Please check and make sure the CI is green. Thanks.

bunnitha:
[0001-Cache-cpu-info-in-a-file-at-boot.patch](https://github.com/user-attachments/files/19403325/0001-Cache-cpu-info-in-a-file-at-boot.patch)
[0001-Trigger-cpuinfo-svc-at-boot.patch](https://github.com/user-attachments/files/19403326/0001-Trigger-cpuinfo-svc-at-boot.patch)
Complete patch to implement the solution in Android platform.

digantdesai (approved):

digantdesai:
Can we make sure CI is green?

digantdesai:
Rebase?

enh-google:
is this different from https://www.phoronix.com/news/Linux-Intel-KVM-Cache-CPUID ? 

when we talked about this at bootcamp, i didn't really mean that you should upstream this to cpuinfo (because i don't think anyone will use it) --- i was just saying "Android won't take non-upstream patches to cpuinfo".

this still feels like something that should be fixed in the kernel, not hacked around in callers...

bunnitha:
> Rebase?

Hi @digantdesai 
The original commit is far behind in the tree and attempts to rebase has messed up the PR.
I have created another one (#304).
Could you please continue the review there? (this one needs to be abandoned)
https://github.com/pytorch/cpuinfo/pull/304

PR #262, opened 2024-09-24
Add `sme2` detection for `aarch64`

gonnet:
Add `cpuinfo_has_arm_sme2` for `aarch64`.

Implemented using the values used [here](https://gitlab.arm.com/kleidi/kleidiai/-/blob/main/test/common/cpu_info.cpp?ref_type=heads#L30).

This also required changing the type of the variables used to store/pass the value of `hwcap2` from `uint32_t` to `uint64_t`, since the flag for `sme2` is at bit 37.


include/cpuinfo.h
```diff
@@ -1671,6 +1671,7 @@ struct cpuinfo_arm_isa {
 	bool sve2;
 	bool i8mm;
 	bool sme;
+        bool sme2;
```
digantdesai, line 1674: tab vs. spaces?

src/arm/linux/api.h
```diff
@@ -138,6 +138,7 @@ struct cpuinfo_arm_linux_proc_cpuinfo_cache {
 #define CPUINFO_ARM_LINUX_FEATURE2_RNG UINT32_C(0x00010000)
 #define CPUINFO_ARM_LINUX_FEATURE2_BTI UINT32_C(0x00020000)
 #define CPUINFO_ARM_LINUX_FEATURE2_SME UINT32_C(0x00800000)
+#define CPUINFO_ARM_LINUX_FEATURE2_SME2 UINT64_C(0x0000002000000000)
```
digantdesai, line 141: Might as well add [others](https://github.com/torvalds/linux/blob/abf2050f51fdca0fd146388f83cddd95a57a008d/arch/arm64/include/uapi/asm/hwcap.h#L99-L104) SME[2] fields?

digantdesai (approved):
LGTM, please fix some nits I commented. I can merge it once you update. Thanks.

src/arm/linux/api.h
gonnet: I've added the bits, should I also add them to `cpuinfo_isa`, with the corresponding `cpuinfo_has_arm_*` functions?

digantdesai: let me know if you are doing this else I will merge this tomorrow.

gonnet: I'll add them!

gonnet: Done :slightly_smiling_face:

gonnet:
This is now good to go!

PR #265, opened 2024-10-27
Disable neon dot  on Unisoc UMS312

GregoryComer:
This PR adds logic to recognize Unisoc UMS312 devices and disable neon dot instructions. This is due to SIGILL crashes on UMS312-based phones. We already disable dot instructions on the Unisoc T310, so this is not unprecedented. It is unclear to me if the UMS312 is actually a SOC or if is a board/platform including the T310. Regardless, it is identified separately, so I have added logic to handle it.

I've also added cpuinfo and ro.props for the Astro 55R, which is a UMS312-based device.

To validate this change, I added tests for matching T301 and UMS312 hardware strings and ran all tests on the Astro 55R. I noticed that there are a large number of existing failing tests, so I re-ran all tests on main and diffed the results to confirm that there are no regressions. I also built XNNPACK with this CPUINFO commit and validated that dot instructions were disabled via observing the relevant log message.


test/build.prop/astro_55r.log
```diff
@@ -0,0 +1,31 @@
+# begin build properties
+# autogenerated by buildinfo.sh
```
digantdesai, line 2: where is this file?

digantdesai (approved):
lgtm, I assume you can trigger the debug log on the real device, right?

src/arm/linux/aarch32-isa.c
```diff
@@ -147,6 +147,8 @@ void cpuinfo_arm_linux_decode_isa_from_proc_cpuinfo(
 				"VDOT instructions disabled: cause occasional SIGILL on Spreadtrum SC9863A");
 		} else if (chipset->series == cpuinfo_arm_chipset_series_unisoc_t && chipset->model == 310) {
 			cpuinfo_log_warning("VDOT instructions disabled: cause occasional SIGILL on Unisoc T310");
+		} else if (chipset->series == cpuinfo_arm_chipset_series_unisoc_ums && chipset->model == 312) {
```
digantdesai, line 150: we you able to repro this on armv8 kernel? And does this fix it? Ok with this change even if you couldn't TBH.

GregoryComer: Unfortunately. I don't have a device with UMS312 and armv8 kernel. From the crash logs, they were mostly (maybe all?) armv7 kernels, so it may be okay. If we see issues on armv8 kernels, I can take as a follow-up.

GregoryComer:
> 

Yes

test/build.prop/astro_55r.log
GregoryComer: It should be located under /system/build.props. However, I can't read it without root, so I pulled the template file and populated this by running adb shell getprop. I can remove this file, if desired.

PR #266, opened 2024-11-14
Detect AMD Zen 5 microarchitecture

fbarchard:
Detect AMD Family 26 with any model as cpuinfo_uarch_zen5


digantdesai (approved):

digantdesai:
From @fbarchard 
```
The reason for wanting zen5 is avx512 on AMD and Intel behave differently
AMD prefers short/wide tile sizes such as 6x64
Intel prefers tall/narrow tile sizes such as 13x32
Also GFNI is present on both, but GFNI is slow on AMD Zen4
And int32 multiply is very slow on Intel but fast on AMD.
```

PR #268, opened 2024-12-06
Set the correct L2 size for Ampere Altra (`aarch64`).

gonnet:
Use the `midr` to identify Ampere Altra cores and set the L2 cache size to 1MiB (as opposed to the default minimum size of 256KiB for `cpuinfo_uarch_neoverse_n1`).

### DIFF
--- a/src/arm/cache.c
+++ b/src/arm/cache.c
@@ -1341,7 +1341,8 @@ void cpuinfo_arm_decode_cache(
 			 * information, please refer to the technical manuals
 			 * linked above
 			 */
-			const uint32_t min_l2_size_KB = uarch == cpuinfo_uarch_neoverse_v2 || midr_is_ampere_altra(midr)
+			const uint32_t min_l2_size_KB =
+				uarch == cpuinfo_uarch_neoverse_v2 || midr_is_ampere_altra(midr)
 					? 1024
 					: 256;
 			const uint32_t min_l3_size_KB = 0;
--- a/src/arm/cache.c
+++ b/src/arm/cache.c
@@ -1341,7 +1341,9 @@ void cpuinfo_arm_decode_cache(
 			 * information, please refer to the technical manuals
 			 * linked above
 			 */
-			const uint32_t min_l2_size_KB = uarch == cpuinfo_uarch_neoverse_v2 ? 1024 : 256;
+			const uint32_t min_l2_size_KB = uarch == cpuinfo_uarch_neoverse_v2 || midr_is_ampere_altra(midr)
+					? 1024
+					: 256;
 			const uint32_t min_l3_size_KB = 0;
 
 			*l1i = (struct cpuinfo_cache){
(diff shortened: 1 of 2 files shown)


src/arm/cache.c
```diff
@@ -1341,7 +1341,8 @@ void cpuinfo_arm_decode_cache(
 			 * information, please refer to the technical manuals
 			 * linked above
 			 */
-			const uint32_t min_l2_size_KB = uarch == cpuinfo_uarch_neoverse_v2 ? 1024 : 256;
+			const uint32_t min_l2_size_KB =
+				(uarch == cpuinfo_uarch_neoverse_v2 || midr_is_ampere_altra(midr)) ? 1024 : 256;
```
digantdesai, line 1345: I guess this is OK, and assume you have verified the sizes somehow. I also assume `CCSIDR_EL1, Current Cache Size ID Register` not readable from EL0.

src/arm/midr.h
```diff
@@ -167,6 +168,11 @@ inline static bool midr_is_kryo_gold(uint32_t midr) {
 	return (midr & uarch_mask) == (CPUINFO_ARM_MIDR_KRYO_GOLD & uarch_mask);
 }
 
+inline static bool midr_is_ampere_altra(uint32_t midr) {
```
digantdesai, line 171: I am not too familiar with Ampere altra line of CPUs, is the function name sufficient to uniquely identify this particular part i.e. do we need some `plus` or `gold` or some suffix :)

digantdesai (approved):

src/arm/midr.h
fbarchard: Altra is based on Cortex-A76 (Pixel4 big core), so as far uarch goes, XNNPack shares kernels that work well on Pixel 4.

But cache will be different

Our 'A75' kernels are actually tuned for A76, since A75 doesnt care much how you load, but A76 does.

The Neoverses are similar
V1 = X1
N1 = A710
V2 = X3
N2 = A720

On Intel I'm doing the same with granite rapids uarch is the same as Meteor lake.  microkernel instruction behavior is sensitive to the uarch.

Knowing the shared uarch makes development easier.  If you want to optimize for Ampere Altra you can test on Pixel 4 or vice versa.

fbarchard (approved):

src/arm/midr.h
gonnet: The Altra Ampere seems to be its own thing, i.e. not a _line_ of CPUs: https://amperecomputing.com/briefs/ampere-altra-family-product-brief.

src/arm/cache.c
gonnet: The Ampere Altra is a `cpu_uarch_neoverse_n1`, i.e. not a new `uarch`.

This function was currently setting the L2 cache size to the known minimum requirement for each `uarch`, regardless of the actual implementation.

gonnet:
If this is good to go, can someone please merge it? Thanks!

PR #269, opened 2024-12-10
Added changes to support AVX 10.1 support

rrwinterton:
detect AVX10.1 ISA

### DIFF
--- a/src/x86/isa.c
+++ b/src/x86/isa.c
@@ -429,6 +429,11 @@ struct cpuinfo_x86_isa cpuinfo_x86_detect_isa(
 	 */
 	isa.avx512f = avx512_regs && !!(structured_feature_info0.ebx & UINT32_C(0x00010000));
 
+    /*
+	 * AVX 10.1 instructions:
+	 */
+    isa.avx10_1 = (structured_feature_info1.edx & UINT32_C(1 << 19));
+	
 	/*
 	 * AVX512PF instructions:
 	 * - Intel: ebx[bit 26] in structured feature info (ecx = 0).
--- a/src/x86/isa.c
+++ b/src/x86/isa.c
@@ -429,10 +429,10 @@ struct cpuinfo_x86_isa cpuinfo_x86_detect_isa(
 	 */
 	isa.avx512f = avx512_regs && !!(structured_feature_info0.ebx & UINT32_C(0x00010000));
 
-    /*
+	/*
 	 * AVX 10.1 instructions:
 	 */
-    isa.avx10_1 = avx512_regs && !!(structured_feature_info1.edx & UINT32_C(0x00080000));
+	isa.avx10_1 = avx512_regs && !!(structured_feature_info1.edx & UINT32_C(0x00080000));
 	
 	/*
 	 * AVX512PF instructions:
--- a/src/x86/isa.c
+++ b/src/x86/isa.c
@@ -432,7 +432,7 @@ struct cpuinfo_x86_isa cpuinfo_x86_detect_isa(
     /*
 	 * AVX 10.1 instructions:
 	 */
-    isa.avx10_1 = (structured_feature_info1.edx & UINT32_C(1 << 19));
+    isa.avx10_1 = avx512_regs && !!(structured_feature_info1.edx & UINT32_C(0x00080000));
 	
 	/*
 	 * AVX512PF instructions:
--- a/src/x86/isa.c
+++ b/src/x86/isa.c
@@ -433,7 +433,7 @@ struct cpuinfo_x86_isa cpuinfo_x86_detect_isa(
 	 * AVX 10.1 instructions:
 	 */
 	isa.avx10_1 = avx512_regs && !!(structured_feature_info1.edx & UINT32_C(0x00080000));
-	
+
 	/*
 	 * AVX512PF instructions:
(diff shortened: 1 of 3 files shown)


src/x86/isa.c
```diff
@@ -429,6 +429,11 @@ struct cpuinfo_x86_isa cpuinfo_x86_detect_isa(
 	 */
 	isa.avx512f = avx512_regs && !!(structured_feature_info0.ebx & UINT32_C(0x00010000));
 
+    /*
+	 * AVX 10.1 instructions:
+	 */
+    isa.avx10_1 = (structured_feature_info1.edx & UINT32_C(1 << 19));
```
fbarchard, line 435: refer a UINT32_C constant like the others
Replace UINT32_C(1 << 19)
with UINT32_C(0x00080000)

To confirm the bits, I look at libyuv AVX10 support

It is in the same register as avxvnniint8
((cpu_einfo7[3] & 0x00000010) ? kCpuHasAVXVNNIINT8 : 0);
((cpu_einfo7[3] & 0x00080000) ? kCpuHasAVX10 : 0) 

It looks right except in libyuv it checks for  if ((GetXCR0() & 0xe0) == 0xe0) which says the OS supports saving ymm16 to ymm31.
So I think you should also check avx512_regs &&

rrwinterton: The reason I didn't check for AVX512 is because in the future there may not be 512 on some systems that support AVX10.2 in 256 bit form but that is still TBD. I think it is a good idea for now since we can change this in the future if this is the case.

fbarchard (approved):
Comment about other avx10 is not needed and subject to change.
The 'avx512 registers' is a misnomer.  The OSS indicates the OS supports saving 32 vector registers.

src/x86/isa.c
```diff
@@ -429,6 +429,11 @@ struct cpuinfo_x86_isa cpuinfo_x86_detect_isa(
 	 */
 	isa.avx512f = avx512_regs && !!(structured_feature_info0.ebx & UINT32_C(0x00010000));
 
+    /*
+	 * AVX 10.1 instructions:
+	 */
+    isa.avx10_1 = avx512_regs && !!(structured_feature_info1.edx & UINT32_C(0x00080000));
```
fbarchard, line 435: indent is wrong.  Make it the same as lines above

fbarchard: to be clear, you should use 1 tab, not 4 spaces

tools/isa-info.c
```diff
@@ -70,6 +70,10 @@ int main(int argc, char** argv) {
 	printf("\tAVX512VP2INTERSECT: %s\n", cpuinfo_has_x86_avx512vp2intersect() ? "yes" : "no");
 	printf("\tAVX512_4VNNIW: %s\n", cpuinfo_has_x86_avx512_4vnniw() ? "yes" : "no");
 	printf("\tAVX512_4FMAPS: %s\n", cpuinfo_has_x86_avx512_4fmaps() ? "yes" : "no");
+	
+	//avx10.1 (in future will add other avx10 version)
```
fbarchard, line 74: this comment is not needed and inconsistent with other printf's

tools/isa-info.c
```diff
@@ -70,6 +70,10 @@ int main(int argc, char** argv) {
...
+	printf("\tAVX10_1: %s\n", cpuinfo_has_x86_avx10_1() ? "yes" : "no");
+
```
fbarchard, line 76: remove blank lines.  avx10_1 is an ISA and should be grouped with the rest

fbarchard (changes requested):
Change title for this PR to Added changes to support AVX 10.1
Have a follow up PR for 10.2

tools/isa-info.c
malfet: CI have linter automation that shoudl detect this

malfet, line 76:
```suggestion
	printf("\tAVX10_1: %s\n", cpuinfo_has_x86_avx10_1() ? "yes" : "no");
```

src/x86/isa.c
```diff
@@ -429,6 +429,11 @@ struct cpuinfo_x86_isa cpuinfo_x86_detect_isa(
 	 */
 	isa.avx512f = avx512_regs && !!(structured_feature_info0.ebx & UINT32_C(0x00010000));
 
+	/*
+	 * AVX 10.1 instructions:
+	 */
+	isa.avx10_1 = avx512_regs && !!(structured_feature_info1.edx & UINT32_C(0x00080000));
+
```
malfet, line 436:
```suggestion
	/*
	 * AVX 10.1 instructions:
	 */
	isa.avx10_1 = avx512_regs && !!(structured_feature_info1.edx & UINT32_C(0x00080000));

```

fbarchard (approved):
Looks good to me.  If you can, test on real hardware that does not have avx512, but does have avx10.1
SDE doesnt really test the OSSAVE portion correctly, and the only publically released hardware that supports avx10.1 is granite rapids, but that also has avx512, so it doesnt test the OSSAVE.
This test is historically so new cpus can be used with old OS's, where the linux kernel does not support the new registers (32 ymm registers), so the cpuid needs to be turned off.
But the way you have it is safe... if it fails, it will just disable avx10 on a cpu/OS that it should have had it enabled.

digantdesai (approved):
Thank you

PR #271, opened 2025-01-11
use _wcsnicmp for string compare which is non-case sensitive

HectorSVC:
use _wcsnicmp for string compare which is non-case sensitive, to avoid the error like:
Error in cpuinfo: Unknown chip model name 'snapdragon (tm) 8cx gen 3 @ 3.40 GHz'.
Please add new Windows on Arm SoC/chip support to arm/windows/init.c!


jywu-msft:
@malfet @ozanMSFT

PR #284, opened 2025-03-21
cmake: rename duplicate names

1480c1:
Xcode on macOS cannot handle multiple .c files having the same name as it produces an object file like `init-<hash>.o`, but tries to link to init.o

https://gitlab.kitware.com/cmake/cmake/-/issues/20501


digantdesai (approved):

malfet:
Reverting as it broke Bazel builds

PR #286, opened 2025-03-27
cmake: convert x86* variants to x86 and x86_64

1480c1:
This should reduce the number of checks for other variants like AMD64, amd64, etc.


digantdesai:
can you rebase? Also for some reason CI didn't run

CMakeLists.txt
```diff
@@ -161,7 +163,7 @@ ENDIF()
 SET(CPUINFO_SRCS src/api.c src/cache.c src/maininit.c src/log.c)
 
 IF(CPUINFO_SUPPORTED_PLATFORM)
-  IF(NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten" AND (CPUINFO_TARGET_PROCESSOR MATCHES "^(i[3-6]86|AMD64|x86(_64)?)$" OR IOS_ARCH MATCHES "^(i386|x86_64)$"))
+  IF(NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten" AND (CPUINFO_TARGET_PROCESSOR MATCHES "^x86(_64)?$" OR IOS_ARCH MATCHES "^(i386|x86_64)$"))
```
fbarchard: should i386 change for ios?  seems like you could remove 32 bit x86 from ios simulator at this point?

fbarchard (approved):
This change makes sense and looks good to me.
x86 and x86_64 are the common names used
would be good to rebase and land it

GregoryComer:
@1480c1 Can you rebase and run clang format? I should be able to merge after that. Thanks!

CMakeLists.txt
1480c1: I don't think that is a decision I should make in this PR. It would be better if a separate PR is created so people who have an interest can comment on it.

1480c1:
> @1480c1 Can you rebase and run clang format? I should be able to merge after that. Thanks!

I didn't touch any C/C++ files, so I don't think there's anything for me to run clang-format on.

PR #283, opened 2025-03-14
Refactor windows arm64 for auto cpu detection

ozanMSFT:
Fixes #270

---

Before:

> Error in cpuinfo: Unknown chip model name 'Snapdragon(R) CPU'.

After fix (tested with `Pytorch 2.7.0` nightly):

> No print - successful detection

Example detection (printed messages only for debugging purposes):

image

---

Refactoring Windows ARM64 CPU detection with the MIDR implementation. 

In this way, Windows detection will use the same structure as Linux.

Previously, some pre-defined CPU information was stored hardcoded in the source code. Now, it is automatically reading MIDR and decoding it.

[MIDR documentation](https://developer.arm.com/documentation/100442/0100/register-descriptions/aarch32-system-registers/midr--main-id-register)

Value is stored in:
> HKEY_LOCAL_MACHINE\HARDWARE\DESCRIPTION\System\CentralProcessor\0\CP 4000

```
CP 4000: MIDR_EL1
CP 4020: ID_AA64PFR0_EL1
CP 4021: ID_AA64PFR1_EL1
CP 4028: ID_AA64DFR0_EL1
CP 4029: ID_AA64DFR1_EL1
CP 402C: ID_AA64AFR0_EL1
CP 402D: ID_AA64AFR1_EL1
CP 4030: ID_AA64ISAR0_EL1
CP 4031: ID_AA64ISAR1_EL1
CP 4038: ID_AA64MMFR0_EL1
CP 4039: ID_AA64MMFR1_EL1
CP 403A: ID_AA64MMFR2_EL1
```
---

Missing [Snapdragon Oryon](https://en.wikipedia.org/wiki/Oryon) is added.

---

Previously `api.h` and `uarch.c` were not used by Windows detection logic. However, to make it common with Linux, we are now using them.

Since `restrict static` is not supported by MSVC, I had to add following macro. 

```cpp
#ifdef _MSC_VER
#define RESTRICT_STATIC /* nothing for MSVC */
#else
#define RESTRICT_STATIC restrict static
#endif
```

PR #289, opened 2025-04-22
Apply `RESTRICT_STATIC` MSVC workaround to `arm/cache.c` as well

MarijnS95:
In https://github.com/pytorch/cpuinfo/pull/283 a workaround was introduced to remove the `restrict static` attributes on MSVC where it is not supported.

This was only defined and used in `arm/api.h` however and not in `arm/cache.c`, leading to this file to fail compilation because it is included in `ARM_SRCS` for the `windows_arm64` target in `BUILD.bazel`.

Fix that by replacing `restrict static` with `RESTRICT_STATIC` which will be defined by the `arm/api.h` header that is already included in `arm/cache.c`.

### DIFF
--- a/src/arm/cache.c
+++ b/src/arm/cache.c
@@ -10,13 +10,13 @@ void cpuinfo_arm_decode_cache(
 	enum cpuinfo_uarch uarch,
 	uint32_t cluster_cores,
 	uint32_t midr,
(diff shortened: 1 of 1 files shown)


digantdesai:
Please sign the CLA to merge the PR

PR #292, opened 2025-05-07
riscv-hw.c: match kernel type in syscall().

enh-google:
syscall() doesn't care about these types at all, and the kernel uses cpu_set_t, so we're better off just removing the cast entirely.

### DIFF
--- a/src/riscv/linux/riscv-hw.c
+++ b/src/riscv/linux/riscv-hw.c
@@ -112,7 +112,7 @@ void cpuinfo_riscv_linux_decode_vendor_uarch_from_hwprobe(
 	 *
 	 * for more details.
 	 */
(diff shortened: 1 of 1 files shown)


digantdesai (approved):

enh-google:
the build failure seems to be a configuration issue (missing riscv64 ubuntu image?), and the clang format errors are for a different file?

malfet:
> the build failure seems to be a configuration issue (missing riscv64 ubuntu image?), and the clang format errors are for a different file?

Parts of the reason it was added were to have some validation of those RISC-V changes, let's me try to fix those

malfet (approved):
Sure, now that check on trunk is green, will revert it it fails them (as I can not rebase)

PR #305, opened 2025-06-30
Add Intel Tremont uarch

fbarchard:
Tremont is the e-core (10nm Atom) in Jasperlake,Elkhart Lake, Jacobville and Lakefield

The family/model are supported by linux
https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/intel-family.h#L176-L178 Documented on wikichip
https://en.wikichip.org/wiki/intel/cpuid#Small_Cores Supported by Intel sde -tnt

Reordered e-core uarch by generation
darkmont is an e-core used in both hybrid laptops and server, so move to ecore gracemont model 0xBE is confirmed for Alderlake N, so redocument support

Predecessor Goldmont Plus
Successor Gracemont

### DIFF
--- a/include/cpuinfo.h
+++ b/include/cpuinfo.h
@@ -373,10 +373,14 @@ enum cpuinfo_uarch {
 	cpuinfo_uarch_goldmont = 0x00100404,
 	/** Intel Goldmont Plus microarchitecture (Gemini Lake). */
 	cpuinfo_uarch_goldmont_plus = 0x00100405,
-	/** Intel Gracemont microarchitecture (Twin Lake). */
-	cpuinfo_uarch_gracemont = 0x00100406,
+	/** Intel Airmont microarchitecture (10 nm out-of-order Atom). */
+	cpuinfo_uarch_tremont = 0x00100406,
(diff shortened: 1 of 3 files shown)


fbarchard:
To test I first run a different command line tool that shows family/model using sde -tnt
Cpu Family 6 (0x6), Model 150 (0x96)
Has SSE2 0x100
Has SSSE3 0x200
Has SSE4.1 0x400
Has SSE4.2 0x800
Has AVX 0x0
Has AVX2 0x0

The model 0x96 is among 3 supported supported by Linux
On wikichip those same 3 are documented, as well as a 4th, which appears to be valid and unique, so all 4 are added.

sde -tnt -- ./cpu-info 
Packages:
	0: Intel Xeon X5680
	1: Intel Xeon X5680
Microarchitectures:
	36x Tremont
Cores:
	0: 37 processors (0-36), Intel Tremont

rrwinterton (approved):
The models and family for the microarchitectures look like the correct values for the new CPUs that are released/being release for Intel CPU's.

PR #307, opened 2025-07-07
Fix Unisoc T618 Chipset Detection

### LINKED ISSUE #306: Chipset Vendor Detection Failure on Unisoc T618
**Device:** Samsung Galaxy Tab A8 (2021, SM-X205N)
**Reference:** [GSMArena - Galaxy Tab A8 (2021)](https://www.gsmarena.com/samsung_galaxy_tab_a8_10_5_%282021%29-11265.php)
**Chipset:** Unisoc Tiger T618

When running `cpuinfo`, the [match_t](https://github.com/pytorch/cpuinfo/blob/e4cadd02a8b386c38b84f0a19eddacec3f433baa/src/arm/linux/chipset.c#L919) function attempts to match the first 8 characters of the CPU name (`Unisoc T`) and then parse the next 3 or 4 characters as digits to determine the model number. However, on this device, the CPU name appears as `UNISOC T618` (note the uppercase `"UNISOC"`), causing the match to fail due to case sensitivity.

**Proposed Solution:**
Update the `match_t` function to perform a case-insensitive comparison when matching the CPU name prefix, ensuring that names like `UNISOC T618` are correctly detected.

bc-lee:
Add support for uppercase "UNISOC T" prefix in `match_t` function to handle devices like Samsung Galaxy Tab A8 (SM-X205N) that report "UNISOC T618" instead of the expected mixed-case "Unisoc T618".

The function now explicitly matches both variants:
- "Unisoc T" (mixed case, existing)
- "UNISOC T" (uppercase, new)

This ensures proper chipset vendor detection on affected Samsung devices where the uppercase variant caused match failures.

Includes test case for "UNISOC T618" detection.
Fixes #306

### DIFF
--- a/src/arm/linux/chipset.c
+++ b/src/arm/linux/chipset.c
@@ -953,7 +953,7 @@ static bool match_sc(const char* start, const char* end, struct cpuinfo_arm_chip
 }
 
 /**
- * Tries to match, case-sentitively, /Unisoc T\d{3,4}/ signature for Unisoc T
+ * Tries to match, case-sentitively, /Unisoc T\d{3,4}/ or /UNISOC T\d{3,4}/ signature for Unisoc T
  * chipset. If match successful, extracts model information into \p chipset
  * argument.
  *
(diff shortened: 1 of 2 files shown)


fbarchard (approved):
Change looks okay and consistent with documentation I find on Galaxy Tab A8

With this change does T618 fill in chipset->model correctly?
Does the T618 require any ISA override?  Example vudot is problematic on T310
https://github.com/pytorch/cpuinfo/blob/main/src/arm/linux/aarch32-isa.c#L148-L151

fbarchard (approved):

GregoryComer:
@bc-lee Can you rebase your PR? I'll get it merged shortly. Thanks!

PR #311, opened 2025-08-01
Add aarch64 to the windows arm list

ykhrustalev:
When cross compiling with llvm on windows, the target cpu is set to aarch64.


CMakeLists.txt
```diff
@@ -184,7 +184,7 @@ IF(CPUINFO_SUPPORTED_PLATFORM)
     ELSEIF(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
       LIST(APPEND CPUINFO_SRCS src/x86/freebsd/init.c)
     ENDIF()
-  ELSEIF(CMAKE_SYSTEM_NAME MATCHES "^Windows" AND CPUINFO_TARGET_PROCESSOR MATCHES "^(ARM64|arm64)$")
+  ELSEIF(CMAKE_SYSTEM_NAME MATCHES "^Windows" AND CPUINFO_TARGET_PROCESSOR MATCHES "^(ARM64|arm64|aarch64)$")
```
fbarchard, line 187: Is that a bug, where ARM64 and arm64 are not the right values to check?  Should it always be aarch64?

kimishpatel (approved):

PR #315, opened 2025-08-05
Add basic NetBSD support.

0-wiz-0:
Based on FreeBSD support.

### DIFF
--- a/src/freebsd/topology.c
+++ b/src/freebsd/topology.c
@@ -52,7 +52,7 @@ struct cpuinfo_freebsd_topology cpuinfo_freebsd_detect_topology(void) {
 		.threads_per_core = 0,
 		.threads = 0,
 	};
-#if defined (__FreeBSD__)
+#if defined(__FreeBSD__)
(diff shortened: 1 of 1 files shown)


src/freebsd/topology.c
```diff
@@ -94,6 +97,19 @@ struct cpuinfo_freebsd_topology cpuinfo_freebsd_detect_topology(void) {
...
+#elif defined(__NetBSD__)
+	topology.cores = sysctl_int("hw.ncpu");
+	if (topology.cores == 0) {
+		goto fail;
+	}
+	cpuinfo_log_debug("netbsd topology: cores = %d", topology.cores);
+	/* more details need active identification of the CPU, no sysctl available */
+	/* TODO: improve */
+	topology.threads = topology.cores;
+	topology.packages = 1;
```
fbarchard, line 109: Can this use the FreeBSD method instead?

0-wiz-0:
No, the sysctl's that FreeBSD has in this area just do not exist on NetBSD. See https://man.netbsd.org/sysctl.7

PR #321, opened 2025-08-16
Disable neon dot on all UNISOC-based phones.

### LINKED ISSUE #322: dot product disabled on aarch32
This change causes a performance regression on arm due to this PR
https://github.com/pytorch/cpuinfo/pull/300

src/arm/linux/aarch32-isa.c ====
152,153d151
< 		} else if (chipset->vendor == cpuinfo_arm_chipset_vendor_unknown) {
< 			cpuinfo_log_warning("VDOT instructions disabled: unknown chipset");

chipset->vendor is unknown on many, if not most new SoC, on all cpus, including qualcomm, google tensor, exynos.

The problem the code atttempts to address is some unisoc/mediatek soc using a faulty linux kernel that marks udot instructions as illegal on Cortex A55.  sdot is okay.  aarch64 udot is okay.  As far as I know, sudot, usdot and i8mm are okay.
Other vendors either used a newer linux kernel, or patched the bug.

To avoid the issue, XNNPack doesnt use udot in either aarch32 or aarch64.  But there is only 1 flag to detect both, and this isa->dot 

Impact is a performance regression
On Samsung S23 qualcomm medium core (A720)
Was QS8MobileNetV2/real_time       10695 us
Now QS8MobileNetV2/process_time/real_time       32577 us

Samsung S23 Qualcomm
SoC name: Unknown
Microarchitectures:
	1x Cortex-X3
	4x Cortex-A715
	3x Cortex-A510
Cores:
	0: 1 processor (0), ARM Cortex-X3
	1: 1 processor (1), ARM Cortex-A715
	2: 1 processor (2), ARM Cortex-A715
	3: 1 processor (3), ARM Cortex-A715
	4: 1 processor (4), ARM Cortex-A715
	5: 1 processor (5), ARM Cortex-A510
	6: 1 processor (6), ARM Cortex-A510
	7: 1 processor (7), ARM Cortex-A510

Samsung S22 Exynos
SoC name: Unknown
Microarchitectures:
	1x Cortex-X2
	3x Cortex-A710
	4x Cortex-A510
Cores:
	0: 1 processor (0), ARM Cortex-X2
	1: 1 processor (1), ARM Cortex-A710
	2: 1 processor (2), ARM Cortex-A710
	3: 1 processor (3), ARM Cortex-A710
	4: 1 processor (4), ARM Cortex-A510
	5: 1 processor (5), ARM Cortex-A510
	6: 1 processor (6), ARM Cortex-A510
	7: 1 processor (7), ARM Cortex-A510

fbarchard:
Instead of disabling neon dot on all unknown vendors, limit the disabling to known unisoc vendor.
Fixes a performance regression seen on Samsung and Pixel devices that have 'unknown' vendor.

On Samsung S23  Qualcomm medium core (A715)
SoC name: Unknown
Microarchitectures:
	1x Cortex-X3
	4x Cortex-A715
	3x Cortex-A510

XNNPACK/bench/subgraph/benchmark --benchmark_filter=MobileNetV2
Was
QS8MobileNetV2/process_time/real_time       32116 us        31926 us           22 cpufreq=2.8032G
Now
QS8MobileNetV2/process_time/real_time       10531 us        10467 us           66 cpufreq=2.8032G

Fixes #322

### DIFF
--- a/src/arm/linux/aarch32-isa.c
+++ b/src/arm/linux/aarch32-isa.c
@@ -142,15 +142,10 @@ void cpuinfo_arm_linux_decode_isa_from_proc_cpuinfo(
 		 * - Neoverse V1 cores
 		 * - Neoverse V2 cores
 		 */
-		if (chipset->series == cpuinfo_arm_chipset_series_spreadtrum_sc && chipset->model == 9863) {
-			cpuinfo_log_warning(
-				"VDOT instructions disabled: cause occasional SIGILL on Spreadtrum SC9863A");
-		} else if (chipset->series == cpuinfo_arm_chipset_series_unisoc_t && chipset->model == 310) {
-			cpuinfo_log_warning("VDOT instructions disabled: cause occasional SIGILL on Unisoc T310");
-		} else if (chipset->series == cpuinfo_arm_chipset_series_unisoc_ums && chipset->model == 312) {
-			cpuinfo_log_warning("VDOT instructions disabled: cause occasional SIGILL on Unisoc UMS312");
-		} else if (chipset->vendor == cpuinfo_arm_chipset_vendor_unknown) {
-			cpuinfo_log_warning("VDOT instructions disabled: unknown chipset");
+		if (chipset->series == cpuinfo_arm_chipset_series_spreadtrum_sc ||
+			chipset->series == cpuinfo_arm_chipset_series_unisoc_t ||
+			chipset->series == cpuinfo_arm_chipset_series_unisoc_ums) {
+			cpuinfo_log_warning("VDOT instructions disabled: cause occasional SIGILL on Unisoc");
 		} else {
 			switch (midr & (CPUINFO_ARM_MIDR_IMPLEMENTER_MASK | CPUINFO_ARM_MIDR_PART_MASK)) {
 				case UINT32_C(0x4100D0B0): /* Cortex-A76 */


kimishpatel (approved):
@GregoryComer this looks good to me? Are you ok with this?

kimishpatel:
> > @GregoryComer this looks good to me? Are you ok with this?
> 
> The chipset series field may not be populated in cases where the chipset is unknown. We should double check this to make sure it doesn't cause the UNISOC crashes to start again. I'm okay to merge this PR to unblock Samsung and Pixel devices, but we'll need to make a follow-up patch, if necessary, before pulling into fbsource.
> 
> I can take a look at the detection failure in more detail tomorrow to see if it would cause the series field to be empty.

Lets validate then before landing

GregoryComer:
> > > @GregoryComer this looks good to me? Are you ok with this?
> > 
> > 
> > The chipset series field may not be populated in cases where the chipset is unknown. We should double check this to make sure it doesn't cause the UNISOC crashes to start again. I'm okay to merge this PR to unblock Samsung and Pixel devices, but we'll need to make a follow-up patch, if necessary, before pulling into fbsource.
> > I can take a look at the detection failure in more detail tomorrow to see if it would cause the series field to be empty.
> 
> Lets validate then before landing

It looks like chipset series will not be populated when the detection fails. Here's the line that's hit on the affected devices:
https://github.com/pytorch/cpuinfo/blob/8a9210069b5a37dd89ed118a783945502a30a4ae/src/arm/linux/chipset.c#L4009-L4016

It bails out of detection and leaves the chipset series as unknown in this case. From a Meta-internal patch, I determined that the specific failure happens because it sees conflicting UNISOC and Spreadtrum vendors. To unblock this, we could probably update the detection logic to special case this. We could also try to use some other field to detect Unisoc in the feature detection logic, instead of the detected chipset, but this seems like it would get messy. I'm definitely open to better ideas.

GregoryComer:
Will Replacing with https://github.com/pytorch/cpuinfo/pull/342 resolve this?

PR #335, opened 2025-10-21
Add missing ARM-Cortex uarchs

tonybaloney:
Added detection for `Cortex-A720`, `Cortex-X4`, `Cortex-X925`, and `Cortex-A725`


tonybaloney:
I'll have to get the datasheets for those and raise in another PR

PR #333, opened 2025-10-20
Implement missing ISA feature flags for ARM64/Windows

tonybaloney:
Most of the ARM64 feature flags aren't being enumerated in Windows.

I haven't found a solution for some advanced SME flags, although no commercial SoC exists for these and Windows doesn't support them yet anyway.

	// TODO: There are no PF flags for these yet. 
	// - sme_i16i32
	// - sme_bi32i32

Also, SVE length is not available outside of the kernel register, so I've set a conservative value of the minimum supported (128 bits).

I tested this on two different ARM64 chipsets and compared it with the reported ISA features in Linux. 

I've copied the Windows SDK values to the header incase the person compiling doesn't have the latest Windows SDK. Some of these values aren't documented. 

The latest list of processor features is:

```c
#define PF_FLOATING_POINT_PRECISION_ERRATA           0   
#define PF_FLOATING_POINT_EMULATED                   1   
#define PF_COMPARE_EXCHANGE_DOUBLE                   2   
#define PF_MMX_INSTRUCTIONS_AVAILABLE                3   
#define PF_PPC_MOVEMEM_64BIT_OK                      4   
#define PF_ALPHA_BYTE_INSTRUCTIONS                   5   
#define PF_XMMI_INSTRUCTIONS_AVAILABLE               6   
#define PF_3DNOW_INSTRUCTIONS_AVAILABLE              7   
#define PF_RDTSC_INSTRUCTION_AVAILABLE               8   
#define PF_PAE_ENABLED                               9   
#define PF_XMMI64_INSTRUCTIONS_AVAILABLE            10   
#define PF_SSE_DAZ_MODE_AVAILABLE                   11   
#define PF_NX_ENABLED                               12   
#define PF_SSE3_INSTRUCTIONS_AVAILABLE              13   
#define PF_COMPARE_EXCHANGE128                      14   
#define PF_COMPARE64_EXCHANGE128                    15   
#define PF_CHANNELS_ENABLED                         16   
#define PF_XSAVE_ENABLED                            17   
#define PF_ARM_VFP_32_REGISTERS_AVAILABLE           18   
#define PF_ARM_NEON_INSTRUCTIONS_AVAILABLE          19   
#define PF_SECOND_LEVEL_ADDRESS_TRANSLATION         20   
#define PF_VIRT_FIRMWARE_ENABLED                    21   
#define PF_RDWRFSGSBASE_AVAILABLE                   22   
#define PF_FASTFAIL_AVAILABLE                       23   
[...]

PR #348, opened 2025-11-21
Add fallback for ARM Windows fp16 detection.

edgchen1:
Background: On a Windows ARM system, I observed that `cpuinfo_has_arm_fp16_arith()` started to return false after upgrading to a more recent cpuinfo version.

In #333, the initialization of `cpuinfo_isa.fp16arith` was updated to use `IsProcessorFeaturePresent(PF_ARM_V82_FP16_INSTRUCTIONS_AVAILABLE)`. I suspect that this is not supported on older Windows versions.

This change adds a fallback path to set `cpuinfo_isa.fp16arith` the old way.
https://github.com/pytorch/cpuinfo/blob/d3a86a813e2bb49d1eb5841ec12e2b135867ab98/src/arm/windows/init.c#L205-L208

### DIFF
--- a/src/arm/windows/init.c
+++ b/src/arm/windows/init.c
@@ -249,6 +249,14 @@ static void set_cpuinfo_isa_fields(void) {
 	// guarantee that, but it holds in practice.
 	cpuinfo_isa.rdm = dotprod;
 
+	// PF_ARM_V82_FP16_INSTRUCTIONS_AVAILABLE may not be available in older
+	// Windows versions. If fp16arith was not detected with
+	// IsProcessorFeaturePresent(PF_ARM_V82_FP16_INSTRUCTIONS_AVAILABLE), fall
+	// back to using the value of dotprod.
+	if (!cpuinfo_isa.fp16arith) {
+		cpuinfo_isa.fp16arith = dotprod;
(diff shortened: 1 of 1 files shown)


fbarchard (approved):
I notice src/arm/mach/init.c has a similar issue with older OS not supporting the fp16 detect, so if the cpu is known it sets it

cpuinfo_isa.fp16arith = get_sys_info_by_name("hw.optional.arm.FEAT_FP16") != 0;
if (!cpuinfo_isa.fp16arith) {
	// Optional in ARMv8.2-A (implemented in Apple cores),
	// list only cores released before iOS 15 / macOS 12
	switch (cpu_family) {
		case CPUFAMILY_ARM_MONSOON_MISTRAL:
		case CPUFAMILY_ARM_VORTEX_TEMPEST:
		case CPUFAMILY_ARM_LIGHTNING_THUNDER:
		case CPUFAMILY_ARM_FIRESTORM_ICESTORM:
			cpuinfo_isa.fp16arith = true;
	}
}

fbarchard (approved):
LGTM  I do think this is always the case and fp16 does not always detect, especially in 32 bit arm.

PR #353, opened 2025-12-08
Adding Zig build

Kbz-8:
I added a Zig build script because I wanted to use cpuinfo in a Zig project. Here's a small PR adding it if you want.

It adds possibility to use this library in Zig codes but also permit to compile the library with the Zig toolchain even if it is not used in Zig.

The build script is mainly based on the BAZEL one.

I do not expect it to be merged as it is a pretty niche adding but I'm opening this PR just in case.

### DIFF
--- a/.github/workflows/build_zig.yml
+++ b/.github/workflows/build_zig.yml
@@ -12,7 +12,7 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 jobs:
-  bazel-linux-local:
+  zig-linux-local:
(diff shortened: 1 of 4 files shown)


fbarchard:
Can you clarify if this is for the zig compiler or a zig build tool?
If it can be treated as a compiler, the main build files (bazel) may be able to use the existing build files?
Also for the compiler aspect, it may be interesting to explore via godbolt, as a way to test code builds correctly with zig

Kbz-8:
This is for the Zig build system and thus will use the Zig compiler. This is mainly for Zig projects wanting to use cpuinfo

PR #355, opened 2025-12-12
Upgrade C++ standard from CXX14 to CXX17

atupone:
It is required by gtest (newest version)


CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ ENDMACRO()
 
 MACRO(CPUINFO_TARGET_ENABLE_CXX11 target)
```
malfet: This macro should have been renamed a while ago

malfet (approved):

PR #375, opened 2026-03-03
Add riscv half-precision floating point detection

ken-unger:
Add cpuinfo_has_riscv_zfh() and cpuinfo_has_riscv_zvfh for fp16 detection.  

The motivation here is to enable this runtime detection support in xnnpack for its rvv fp16 kernels. (xnnpack uses this library)


fbarchard (approved):
LG.  can you confirm this is vector arithmetics and is there a different detect for conversion?
another isa that would be great to detect is the 8 bit dot product

fbarchard (approved):

ken-unger:
Is there a maintainer who could review and merge (if accepted) this PR?  Thank you.

fbarchard:
As we are waiting on this to land, and then roll into XNNPack, I've added similar code in hardware-config.c for XNNPack.
https://github.com/google/XNNPACK/pull/9892

It should be removed in favor of cpuinfo when possible.

malfet (approved):

PR #384, opened 2026-04-28
Add Cortex-A320 to MIDR decode table

npitre:
Split out from #379 per review request.

ARM Cortex-A320 (MIDR part `0xD8F`) is an ARMv9.2-A efficiency core.
Add its uarch enum value and MIDR decode entry so consumers (XNNPACK,
KleidiAI, etc.) can dispatch optimised kernels when running on this
core.

The A320 implements the ARMv9.2-A mandatory feature set: NEON, SVE2,
dotprod, FP16, BF16, I8MM (per the [Cortex-A320 TRM][trm]).

The new MIDR/uarch entries are inserted in numerical order alongside
the existing ARMv9 cores added by recent commits (A520, A720, X4,
X925, A725, Lumex variants).

[trm]: https://developer.arm.com/Processors/Cortex-A320

### DIFF
--- a/include/cpuinfo.h
+++ b/include/cpuinfo.h
@@ -525,6 +525,8 @@ enum cpuinfo_uarch {
 	cpuinfo_uarch_cortex_a510 = 0x00300551,
 	/** ARM Cortex-A520. */
 	cpuinfo_uarch_cortex_a520 = 0x00300552,
+	/** ARM Cortex-A320. */
+	cpuinfo_uarch_cortex_a320 = 0x00300553,
 	/** ARM Cortex-A710. */
 	cpuinfo_uarch_cortex_a710 = 0x00300571,
 	/** ARM Cortex-A715. */
--- a/src/arm/uarch.c
+++ b/src/arm/uarch.c
@@ -141,17 +141,20 @@ void cpuinfo_arm_decode_vendor_uarch(
 				case 0xD87: /* Cortex-A725 */
 					*uarch = cpuinfo_uarch_cortex_a725;
 					break;
-				case 0xD8C:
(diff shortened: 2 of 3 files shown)


fbarchard (approved):
once landed i'll roll it into xnnpack so the uarch_cortex_a320 is available.  or youre welcome to, but its almost easier for me to do it.  If youre able to run benchmarks in xnnpack, you could select kernels that really are faster on a320.  likely a55 or whatever a520 is doing.

npitre:
Thanks for the approval!

Happy to let you handle the XNNPACK side — appreciate the offer. Mapping to `xnn_uarch_cortex_a510` would be a sensible interim choice (closest existing entry); we can revisit once silicon lands. We also have an XNNPACK PR open ([#10060][pr10060]) adding generic Zephyr platform support, which is what surfaced the A320 question in the first place.

On benchmarks: not yet, unfortunately — we're running on the Corstone-1000-A320 FVP, which is functionally accurate but not cycle-accurate for the CPU side, so any kernel ordering we'd derive from it would be misleading. Once we have real A320 silicon we'll come back with measurements and can compare against A55 / A520 kernel paths.

[pr10060]: https://github.com/google/XNNPACK/pull/10060

fbarchard:
Also update ./tools/cpu-info.c with the string.  Ideally build and run the tool to confirm detects.

npitre:
Done — added the `Cortex-A320` string mapping in `tools/cpu-info.c` and confirmed the tool builds. The Zephyr backend's runtime detection on the A320 FVP correctly resolves to `cpuinfo_uarch_cortex_a320` (verified via XNNPACK's `cpu_get_uarch` reporting `0x300553` once #379 is in place).

PR #389, opened 2026-05-16
Normalization of AMD brand string with "with Radeon Graphics"

johnthacker:
Some AMD processors with integrated graphics have "with Radeon Graphics" in the cpuid brand string without being preceded by other tokens like "SOC" or "APU" that are already used to erase the string. With the current logic, everything starting with "Radeon" is erased, leading to less than ideal results. Add a check for a "with" token and erase everything starting with that.

Running cpu-info before:

Packages:
        0: AMD Ryzen 5 PRO 6650U with

and after:

Packages:
        0: AMD Ryzen 5 PRO 6650U

There are plenty of other examples at
https://github.com/InstLatx64/InstLatx64/tree/master/AuthenticAMD

### DIFF
--- a/src/x86/name.c
+++ b/src/x86/name.c
@@ -307,6 +307,15 @@ static bool transform_token(char* token_start, char* token_end, struct parser_st
 			}
 			break;
 		case 4:
+			/*
+			 * Erase everything starting with "with" on AMD
+			 * processors, e.g. "AMD Ryzen 5 PRO 6650U with Radeon
+			 * Graphics"
+			 */
+			if (erase_matching(token_start, token_length, "with")) {
+				return false;
+			}
+
 			/* Remember to erase "Dual Core" in "AMD Athlon(tm) 64
 			 * X2 Dual Core Processor 3800+" */
 			if (memcmp(token_start, "Dual", token_length) == 0) {
--- a/test/name/brand-string.cc
+++ b/test/name/brand-string.cc
@@ -328,6 +328,7 @@ TEST(BRAND_STRING, amd) {
 		"Phenom II X6 1055T",
 		normalize_brand_string("AMD Phenom(tm) II X6 1055T Processor\0\0\0\0\0\0\0\0\0\0\0\0"));
 	EXPECT_EQ("Ryzen 5 1500X", normalize_brand_string("AMD Ryzen 5 1500X Quad-Core Processor          \0"));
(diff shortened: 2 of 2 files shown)


fbarchard:
I dont know the erase function(), so I had antigravity review this change

Yes, this PR makes complete sense.

  • The Problem: AMD APUs (processors with integrated graphics) include graphics information in their CPUID brand string, typically starting with  "with Radeon Graphics" .
  • Existing Logic:  cpuinfo  aims to extract only the CPU model (e.g.,  "Ryzen 5 PRO 6650U" ). For older models, it encountered tokens like  "APU"  or  "SOC"  first, which triggered truncation, discarding  "with Radeon Graphics" . However, newer
  AMD processors (like the Ryzen 5 PRO 6650U) do not have  "APU"  or  "SOC"  in their brand string.
  • The Bug: Without  "APU"  or  "SOC" , the parser continued until it hit  "Radeon" , which it stripped along with everything after it. This left the trailing word  "with"  in the normalized string, resulting in  "AMD Ryzen 5 PRO 6650U with" .
  • The Fix: The PR adds  "with"  as a truncating token. When  "with"  is encountered, it is erased, and the parser truncates the string immediately, effectively discarding the graphics information and leaving only the correct CPU model.

  ### 2. Coding Conventions

  The implementation perfectly aligns with  cpuinfo 's existing brand string normalization design in name.c:

  • It integrates into the  transform_token  function under  case 4:  (since  "with"  has 4 characters).
  • It utilizes the helper  erase_matching  to replace  "with"  with spaces and returns  false  to signal the caller to truncate the string.
  • It includes a test case in brand-string.cc to verify the fix against the real-world brand string:  "AMD Ryzen 5 PRO 6650U with Radeon Graphics" .

  ### 3. Brand Strings with "with"

  • In x86 CPUID brand strings,  "with"  is almost exclusively used by AMD to denote integrated graphics (e.g.,  "with Radeon Graphics" ,  "with Radeon(TM) R7 Graphics" ).
  • Intel brand strings do not typically use  "with"  in their CPUID model names.
  • Therefore, treating  "with"  as a truncation trigger is a safe and robust way to strip integrated graphics marketing suffixes from AMD processors without risking accidental truncation of important model information.

fbarchard (dismissed):

fbarchard (dismissed):

GregoryComer (dismissed):

GregoryComer:
CI seems stuck. I'm going to close and re-open to try to unstick it.

GregoryComer (approved):

PR #385, opened 2026-04-28
Add Apple M1/M2/M3/M4 Linux MIDR detection

fbarchard:
Fixes #380 by adding `cpuinfo_arm_decode_vendor_uarch` logic for Apple's implementer `'a'` (0x61).

This PR includes MIDR part mappings for:
- **M1**: Icestorm / Firestorm (0x022, 0x023, 0x024, 0x025, 0x028, 0x029)
- **M2**: Blizzard / Avalanche (0x032, 0x033, 0x034, 0x035, 0x038, 0x039)
- **M3**: Sawtooth / Everest (0x042, 0x043, 0x044, 0x045, 0x048, 0x049)
- **M4**: Donan Sawtooth / Donan Everest (0x052, 0x053)

These additions allow `cpuinfo` to correctly detect the microarchitecture of Apple Silicon cores when running on Linux (e.g., Asahi Linux).

PR #395, opened 2026-06-16
Initialize vendor/uarch out-parameters for unhandled implementers

### LINKED ISSUE #394: cpuinfo_arm_decode_vendor_uarch returns uninitialized vendor/uarch for unhandled implementers
The 'default:' branch in cpuinfo_arm_decode_vendor_uarch() ('src/arm/uarch.c') only logs a warning and never writes '*vendor' / '*uarch'. The Windows-on-Arm caller get_core_info_from_midr() ('src/arm/windows/init.c') passes pointers to uninitialized stack variables, so for any unhandled implementer the resulting vendor/uarch are whatever happens to be on the stack. The garbage then propagates into every 'cpuinfo_core::vendor' via cpu_info_init_by_logical_sys_info().

maxim-davgalev:
Initialize the vendor/uarch out-parameters at function entry in
cpuinfo_arm_decode_vendor_uarch (src/arm/uarch.c), so an unhandled
implementer yields unknown values rather than uninitialized stack values.

Fixes #394

### DIFF
--- a/src/arm/uarch.c
+++ b/src/arm/uarch.c
@@ -11,6 +11,11 @@ void cpuinfo_arm_decode_vendor_uarch(
 #endif /* CPUINFO_ARCH_ARM */
 	enum cpuinfo_vendor vendor[RESTRICT_STATIC 1],
 	enum cpuinfo_uarch uarch[RESTRICT_STATIC 1]) {
+	/* Ensure the out-parameters are always initialized, including for
+	 * implementers that are not handled in the switch below. */
+	*vendor = cpuinfo_vendor_unknown;
+	*uarch = cpuinfo_uarch_unknown;
+
 	switch (midr_get_implementer(midr)) {
 		case 'A':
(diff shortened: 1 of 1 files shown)


fbarchard:
Does https://github.com/pytorch/cpuinfo/pull/385 cover this?

maxim-davgalev:
> Does #385 cover this?

Yes, #385 covers part of this.

#395 addresses two defects in cpuinfo_arm_decode_vendor_uarch:

1. Missing Apple implementer ('a' / 0x61) case. #385 fully covers this.

2. The 'default' branch only logs a warning and never writes the *vendor / *uarch out-parameters, while the caller passes uninitialized stack variables. #385 does not cover this.

So, as #385 fixes logic for Apple's implementer 'a' (0x61), I can drop the Apple-case patch and update the PR and issue #394.

maxim-davgalev:
I've updated the PR and issue #394.

GregoryComer (approved):

PR #413, opened 2026-07-10
Add ARM FP8 detection

### LINKED ISSUE #412: add FP8 detect
Arm has 4 features in neon (8.7)
cvt
fma
fdot
f8mm

For isa_info I propose it expose the type, dot product and f8mm as 3 isa features for neon

        ARM v8.7 FP8: yes
        ARM v8.7 FP8 dot product: yes
        ARM v8.7 FP8 matrix multiplication: yes

Intel could expose fp8 for convert and amx.

fbarchard:
Add ARM v8.7 / Armv9.6-A FP8 instructions across public API structs, Linux HWCAP2 / /proc/cpuinfo parsing, Apple sysctls, Windows processor feature checks, and isa_info:
- Base FP8 (`cpuinfo_has_arm_fp8` / `FEAT_FP8` / `HWCAP2_F8CVT`)
- FP8 4-way Dot Product (`cpuinfo_has_arm_f8dot` / `FEAT_FP8DOT4` / `HWCAP2_F8DP4`)
- FP8 8-way Matrix Multiply-Accumulate into FP32 (`cpuinfo_has_arm_f8mm` / `FEAT_F8F32MM` / `HWCAP_F8MM8` / `HWCAP2_F8FMA`)

Fixes https://github.com/pytorch/cpuinfo/issues/412

### DIFF
--- a/src/arm/linux/api.h
+++ b/src/arm/linux/api.h
@@ -119,6 +119,7 @@ struct cpuinfo_arm_linux_proc_cpuinfo_cache {
 #define CPUINFO_ARM_LINUX_FEATURE_SB UINT32_C(0x20000000)
 #define CPUINFO_ARM_LINUX_FEATURE_PACA UINT32_C(0x40000000)
 #define CPUINFO_ARM_LINUX_FEATURE_PACG UINT32_C(0x80000000)
+#define CPUINFO_ARM_LINUX_FEATURE_F8MM UINT64_C(0x0000000800000000)
 
 #define CPUINFO_ARM_LINUX_FEATURE2_DCPODP UINT32_C(0x00000001)
 #define CPUINFO_ARM_LINUX_FEATURE2_SVE2 UINT32_C(0x00000002)
@@ -145,6 +146,8 @@ struct cpuinfo_arm_linux_proc_cpuinfo_cache {
 #define CPUINFO_ARM_LINUX_FEATURE2_SME_BI32I32 UINT64_C(0x0000010000000000)
 #define CPUINFO_ARM_LINUX_FEATURE2_SME_B16B16 UINT64_C(0x0000020000000000)
 #define CPUINFO_ARM_LINUX_FEATURE2_SME_F16F16 UINT64_C(0x0000040000000000)
+#define CPUINFO_ARM_LINUX_FEATURE2_FP8 UINT64_C(0x0008000000000000)
+#define CPUINFO_ARM_LINUX_FEATURE2_F8DOT UINT64_C(0x0020000000000000)
 #endif
 
 #define CPUINFO_ARM_LINUX_VALID_ARCHITECTURE UINT32_C(0x00010000)
@@ -179,7 +182,7 @@ struct cpuinfo_arm_linux_processor {
 	uint32_t architecture_flags;
 	struct cpuinfo_arm_linux_proc_cpuinfo_cache proc_cpuinfo_cache;
 #endif
-	uint32_t features;
+	uint64_t features;
 	uint64_t features2;
 	/**
 	 * Main ID Register value.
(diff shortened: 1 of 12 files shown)


src/arm/linux/api.h
```diff
@@ -145,6 +145,9 @@ struct cpuinfo_arm_linux_proc_cpuinfo_cache {
 #define CPUINFO_ARM_LINUX_FEATURE2_SME_BI32I32 UINT64_C(0x0000010000000000)
 #define CPUINFO_ARM_LINUX_FEATURE2_SME_B16B16 UINT64_C(0x0000020000000000)
 #define CPUINFO_ARM_LINUX_FEATURE2_SME_F16F16 UINT64_C(0x0000040000000000)
+#define CPUINFO_ARM_LINUX_FEATURE2_FP8 UINT64_C(0x0008000000000000)
+#define CPUINFO_ARM_LINUX_FEATURE2_F8MM UINT64_C(0x0010000000000000)
```
JakeStevens: [Here](https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/hwcap.h) this is HWCAP2_F8FMA eg (1UL << 52).

Should we rename this to F8FMA?

Or use one of these:
#define HWCAP_F8MM8		(1UL << 35)
#define HWCAP_F8MM4		(1UL << 36)

fbarchard: no for 2 reasons
1. we are not planning to use fma and its okay for that feature to not be supported.  (but it is supported)
2. for consistency with BF16 which has a general type.  But my intent is that if FP8 is supported but dot product and matrix multiply are not, then fallback on F8CVT

if you dont like FP8 as a type by itself (like BF16) then explicitely add F8CVT instead would be my preference.  But that is less consistent with BF16 and FP16

JakeStevens: Sorry if I was not clear.

I mean that 

CPUINFO_ARM_LINUX_FEATURE2_F8MM

Is currently set to

UINT64_C(0x0010000000000000)

Which it I understand correctly from my link is actually 

HWCAP2_F8FMA

In the Linux header.

So it seems like the code as is is checking for fma

fbarchard: Done.  Switched to HWCAP_F8MM8 which is beyond the 32 bit int, but we can check directly:
if (hwcap_val & (1UL << 35)) { // HWCAP_F8MM8

src/arm/linux/hwcap.c
```diff
@@ -63,8 +63,12 @@ bool cpuinfo_arm_linux_hwcap_from_getauxval(uint32_t hwcap[restrict static 1], u
 		goto cleanup;
 	}
 
-	*hwcap = getauxval(AT_HWCAP);
+	unsigned long hwcap_val = getauxval(AT_HWCAP);
+	*hwcap = (uint32_t)hwcap_val;
 	*hwcap2 = getauxval(AT_HWCAP2);
+	if ((uint64_t)hwcap_val & (UINT64_C(1) << 35)) {
+		*hwcap2 |= CPUINFO_ARM_LINUX_FEATURE2_F8MM;
```
JakeStevens: We are checking hwcap_val, and then setting our local hwcap2 here.

So, we are basically stuffing something from feature into feature2 in cpuinfo, right?

What if that bit is already set for a different reason, eg

hwcap2_val & (UINT64_C(1) << 35) == true

already, but hwcap_val & (UINT64_C(1) << 35) == false?

We would report FP8MM support but it won't actually have it.

If the problem is the underlying linux feature bit fields are too wide, do we just need to widen the features/features2 fields in cpuinfo here?

fbarchard: widening hwcap to uint64_t is likely the cleaner solution, but a bigger change.

PR #414, opened 2026-07-11
fix off-by-one write in proc/cpuinfo hardware/revision parse

soma0212:
In `parse_line` (src/arm/linux/cpuinfo.c) the Hardware and Revision handlers null-terminate with `state->hardware[value_length] = '\0'` in the branch taken when the value fits. But the length guard uses `> CPUINFO_HARDWARE_VALUE_MAX` (and `> CPUINFO_REVISION_VALUE_MAX`), so a value whose length is exactly the max still takes that branch. Those destinations are exactly MAX bytes (64 and 9), so a `Hardware` line with a 64-byte value, or a `Revision` line with a 9-byte value, writes the terminator one byte past the buffer.

Change both comparisons to `>=` so an equal-length value goes through the truncation path instead. That leaves the buffer full and unterminated, which is already the contract downstream code relies on (the chipset decoders read these fields with `strnlen(buf, MAX)`), so keeping the bound at the write site is enough and no consumer needs to change.

### DIFF
--- a/src/arm/linux/cpuinfo.c
+++ b/src/arm/linux/cpuinfo.c
@@ -872,7 +872,7 @@ static bool parse_line(
 				/* BogoMIPS is useless, don't parse */
 			} else if (memcmp(line_start, "Hardware", key_length) == 0) {
 				size_t value_length = value_end - value_start;
-				if (value_length > CPUINFO_HARDWARE_VALUE_MAX) {
+				if (value_length >= CPUINFO_HARDWARE_VALUE_MAX) {
 					cpuinfo_log_warning(
 						"length of Hardware value \"%.*s\" in /proc/cpuinfo exceeds limit (%d): truncating to the limit",
 						(int)value_length,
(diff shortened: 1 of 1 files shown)


fbarchard (approved):

fbarchard:
You'll need to sign the Contributor License Agreement

mhansen:
Dupe of https://github.com/pytorch/cpuinfo/pull/408

soma0212:
Right, #408 predates this one and fixes the same off-by-one; I missed it when filing. The only behavioral difference is the exactly-max-length case: #408 truncates to 63 bytes and always null-terminates, while this keeps the full 64 bytes unterminated, which the strnlen(hardware, CPUINFO_HARDWARE_VALUE_MAX) consumers in chipset.c already handle. Both close the OOB write, so whichever approach you prefer to land works for me. I'll get the CLA sorted in the meantime.

soma0212:
Will get the CLA signed.

On the dupe, yes, same underlying bug and #408 predates this one. The practical difference is behavior at the limit: #408 truncates to MAX-1 and always null-terminates, so an exactly-64-byte Hardware value loses its last character, while this patch keeps the full value unterminated, which is what the chipset decoders already assume (they read these fields with strnlen(buf, MAX) in chipset.c). Either version closes the OOB write, so I'll defer to whichever approach you want to land, and I'm fine closing this one if you take #408.

PR #415, opened 2026-07-13
Add AVX-VNNI-FP16 ISA feature detection and reporting to cpuinfo

fbarchard:
Expose AMD Zen 6 / Intel AVX-VNNI-FP16 (`avx_vnni_fp16`) instructions in `pytorch/cpuinfo`:
- Add `avx_vnni_fp16` boolean field to `struct cpuinfo_x86_isa`
- Add public inline accessor `cpuinfo_has_x86_avx_vnni_fp16(void)` (guarded for 64-bit mode)
- Detect support via CPUID Leaf 0x7, Sub-leaf 0x1 (ECX = 1), EAX bit 1 (`0x00000002`)
- Report `AVX_VNNI_FP16` support in `isa-info` tool

### DIFF
--- a/src/x86/isa.c
+++ b/src/x86/isa.c
@@ -588,6 +588,12 @@ struct cpuinfo_x86_isa cpuinfo_x86_detect_isa(
 	isa.avx_vnni_int16 = avx_regs && !!(structured_feature_info1.edx & UINT32_C(0x00000400));
 
 	/*
+	 * AVX_VNNI_FP16 instructions:
+	 * - AMD/Intel: eax[bit 1] in structured feature info (ecx = 1).
+	 */
+	isa.avx_vnni_fp16 = avx_regs && !!(structured_feature_info1.eax & UINT32_C(0x00000002));
+
+	/*
 	 * AVX_NE_CONVERT instructions:
 	 * - Intel: edx[bit 5] in structured feature info (ecx = 1).
 	 */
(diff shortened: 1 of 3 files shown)


src/x86/isa.c
```diff
@@ -587,6 +587,12 @@ struct cpuinfo_x86_isa cpuinfo_x86_detect_isa(
 	 */
 	isa.avx_vnni_int16 = avx_regs && !!(structured_feature_info1.edx & UINT32_C(0x00000400));
 
+	/*
+	 * AVX_VNNI_FP16 instructions:
+	 * - AMD/Intel: eax[bit 1] in structured feature info (ecx = 1).
+	 */
+	isa.avx_vnni_fp16 = avx_regs && !!(structured_feature_info1.eax & UINT32_C(0x00000002));
```
JakeStevens, line 594: IIUC this seems to map to:

X86_FEATURE_SM3

https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/cpufeatures.h

fbarchard: Good catch.  SM3 is for SHA-256 crypto.  Its not clear there is a dedicated cpuid for this.
For evex encoding AVX512 FP16 can be used for now.

For avx2 it is unclear if/how AVX-VNNI-FP16 detect will be done.

fbarchard: Although I think we'll want this detect, I dont have enough information to implement it for amd.
Intel side fp16 is part of avx10.2 so we can detect it that way, and if future cpus have avx2 or avx512 dot product we can add it then

PR #418, opened 2026-07-14
fix out-of-bounds l2 cache write in cpuinfo_arm_linux_init

soma0212:
The L2 cache count and population passes in cpuinfo_arm_linux_init disagree on when a core gets an L2 entry:
- counting keys off the decoded L2 size, but population overrides it with the size from sysfs (cache/index2/size)
- a core whose built-in decode reports no L2 (ARM11, or an unrecognized ARMv6 core) but whose sysfs exposes an L2 size is skipped while counting yet still gets a slot while populating
- l2_index then advances past l2_count, writing past the end of the calloc'd l2 array, or through a NULL l2 when the count is zero (the L2 writes are unguarded, unlike the L3 writes)

Applying the same sysfs size override in the counting pass keeps l2_count aligned with the number of entries written.

### DIFF
--- a/src/arm/linux/init.c
+++ b/src/arm/linux/init.c
@@ -806,22 +806,26 @@ void cpuinfo_arm_linux_init(void) {
 			 * between all cores
 			 */
 			shared_l3 = false;
-			if (temp_l2.size != 0) {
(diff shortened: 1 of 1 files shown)


soma0212:
any update?

PR #420, opened 2026-07-15
cast to unsigned char before toupper in cache size parser

soma0212:
toupper() in cache_size_parser gets a plain char, which is signed on the ARM/Android targets this file builds for:
- the argument is the first non-digit byte of the sysfs cache-size value (/sys/.../cache/indexN/size), so any byte >= 0x80 reaches toupper as a negative int
- C11 7.4p1 requires the argument to be representable as unsigned char or equal to EOF; bionic and musl index their ctype table directly with it, so a negative value is an out-of-bounds read
- cast to unsigned char, the standard guard for ctype on possibly-signed input

### DIFF
--- a/src/arm/linux/init.c
+++ b/src/arm/linux/init.c
@@ -46,7 +46,7 @@ static bool cache_size_parser(const char* filename, const char* text_start, cons
 		return false;
 	}
 	uint32_t multiplier = 1024;
-	if (p < text_end && toupper(*p) == 'M') {
+	if (p < text_end && toupper((unsigned char)*p) == 'M') {
 		multiplier = 1024 * 1024;
 	}
 	*size_ptr = value * multiplier;


fbarchard (approved):
makes sense.  its good there is a comment above showing (e.g. "2048K", "1M") 
for context.  Makes me think the default 
uint32_t multiplier = 1024;
should really be '1' and check for 'K' the same way.

or just expect it to be upper case.  when I check my linux workstation it says
more /sys/devices/system/cpu/cpu0/cache/index0/size 
32K
pixel2 android device is
adb shell more /sys/devices/system/cpu/cpu0/cache/index0/size 
32K

soma0212:
Right, the generic cacheinfo driver formats this value as "%uK" (size_show in drivers/base/cacheinfo.c), so on mainline kernels the suffix is always present and always uppercase, which lines up with what you're seeing on both your boxes. That means the KB default and the toupper only matter for vendor kernels that stray from that, and treating a bare number as KB has been this parser's behavior all along. I kept this PR to just the cast so it fixes the UB without changing behavior on any input, but happy to switch to multiplier = 1 with an explicit 'K' check in a follow-up if you'd prefer that shape.

fbarchard:
thanks for the details.  I think its good as is... no reason to risk introducing a bug to avoid toupper.
If you want this landed as you need to sign the Contributor License Agreement.

If you plan to never work on cpuinfo again, you could ask me or someone else to make/land the PR

soma0212:
CLA's signed now.

PR #429, opened 2026-07-29
fix stack out-of-bounds read in core cluster heuristic

soma0212:
cluster_processors[3] is indexed by a counter that nothing bounds:

- the table is sized from usable_processors, which cpuinfo_arm_linux_init passes as valid_processors (VALID flag and listed in /proc/cpuinfo), while both loops start a new cluster for every processor carrying just CPUINFO_LINUX_FLAG_VALID
- the two counts diverge whenever a CPU is in the possible and present lists but absent from /proc/cpuinfo, which init.c already calls fairly common for offline cores
- 12 present CPUs with 10 listed in /proc/cpuinfo puts the 11th at cluster #3 and reads cluster_processors[3]; the 5/6/8 tables never write entry [2] either, so those run into uninitialized stack first
- guarding only the verification pass is deliberate, since the assignment pass walks the same processors in the same order, and returning false from there would leave package_leader_id half rewritten, which this function documents it never does

Exhausting the table now fails the heuristic, so the caller falls back to the sequential scan the way it already does for every other mismatch.


soma0212:
any update?

PR #431, opened 2026-07-31
fix stack overread on truncated range in cpulist parse_entry

soma0212:
parse_entry dereferences number_start without checking that anything actually follows the '-':

- cpuinfo_linux_parse_cpulist hands it entry_end == &buffer[BUFFER_SIZE] whenever a full read turns up no ',', since the leftover then gets reparsed as one entry spanning the whole buffer
- an entry ending in '-' leaves number_start at entry_end, parse_number consumes nothing, and the failure branch reads *number_start, which is buffer[256]
- cpuinfo_log_warning is an inline static variadic function, so that argument is evaluated whatever CPUINFO_LOG_LEVEL is set to
- ASan calls it a 1-byte stack-buffer-overflow read at cpulist.c:119 on a cpu list of 255 digits followed by '-'

Bailing out before the dereference leaves the entry rejected exactly as it already was, so the callback sequence and return value are the same on every input I tried, valid or malformed.


fbarchard:
Do you have an example of this actually occuring?

soma0212:
Not from a stock kernel, no. The cpu lists real sysfs emits are short and comma separated, so I don't have a device in the wild hitting this. I found it while exercising the parser's buffer refill boundary under ASan: the trigger is a cpu list file whose first 256 bytes (BUFFER_SIZE) contain no ',' and end in '-'. In that case the whole buffer is reparsed as a single entry with entry_end == &buffer[256], parse_number consumes nothing after the '-', and the warning branch reads buffer[256]. The read happens at any CPUINFO_LOG_LEVEL because cpuinfo_log_warning is a variadic function, so its arguments are evaluated even when the message is filtered out.

Standalone repro against unpatched main (66ee79c):

```c
#include <stdio.h>
#include <string.h>
#include <stdbool.h>
#include <stdint.h>
#include <linux/api.h>

static bool cb(uint32_t start, uint32_t end, void* ctx) {
	printf("callback: [%u, %u)\n", start, end);
	return true;
}

int main(void) {
	/* 255 digits followed by '-' == exactly BUFFER_SIZE (256) bytes, no comma,
	 * so parse_entry() is called with entry_end == &buffer[256]. */
	char content[257];
	memset(content, '0', 255);
	content[255] = '-';
	content[256] = '\0';

	const char* path = "/tmp/cpuinfo_repro_cpulist.txt";
	FILE* f = fopen(path, "wb");
	fwrite(content, 1, 256, f);
	fclose(f);

	bool ok = cpuinfo_linux_parse_cpulist(path, cb, NULL);
	printf("parse_cpulist -> %d\n", (int)ok);
	return 0;
}
```

```
clang -fsanitize=address -fno-omit-frame-pointer -g -O1 \
  -Iinclude -Isrc -Ideps/clog/include -DCPUINFO_LOG_LEVEL=5 \
  repro_cpulist.c src/linux/cpulist.c src/log.c -o repro_cpulist
./repro_cpulist
```

ASan reports a 1-byte stack-buffer-overflow READ at cpulist.c:119 in parse_entry, with 'buffer' at frame offsets [32, 288) and the access at offset 288. With this patch the same harness runs clean and the entry is rejected with the usual warning, so I'd file it under hardening the parse boundary rather than something a real machine produces today. I also checked the callback sequence and return value stay byte identical across a set of valid and malformed lists ("0-3", "0-3,8-11", "3-1", "5-", "5-,7", etc).

soma0212:
Not from a real kernel, no. The /sys cpu-list files never carry a 255-digit number, so this won't fire on healthy hardware. It's reachable through the parser itself though: when a full 256-byte read turns up no comma, `cpuinfo_linux_parse_cpulist` reparses the whole buffer as a single entry with `entry_end == &buffer[256]`. If that entry ends in `-`, `number_start` lands on `entry_end`, `parse_number` consumes nothing, and the failure branch reads `*number_start`, which is `buffer[256]`, one past the buffer. `cpuinfo_log_warning` is an inline static variadic, so that argument is evaluated regardless of `CPUINFO_LOG_LEVEL`.

Smallest trigger is a cpu-list file of exactly 256 bytes with no comma and a trailing `-` (255 digits then `-`). Standalone ASan harness against the current source:

```c
#include <stdio.h>
#include <string.h>
#include <stdbool.h>
#include <stdint.h>
#include <linux/api.h>

static bool cb(uint32_t start, uint32_t end, void* ctx) {
	printf("callback: [%u, %u)\n", start, end);
	return true;
}

int main(void) {
	/* 255 digits + '-' == exactly BUFFER_SIZE (256) bytes, no comma,
	 * so parse_entry() runs with entry_end == &buffer[256]. */
	char content[256];
	memset(content, '0', 255);
	content[255] = '-';

	const char* path = "/tmp/cpuinfo_repro_cpulist.txt";
	FILE* f = fopen(path, "wb");
	fwrite(content, 1, 256, f);
	fclose(f);

	printf("parse_cpulist -> %d\n", (int)cpuinfo_linux_parse_cpulist(path, cb, NULL));
	return 0;
}
```

```
clang -fsanitize=address -fno-omit-frame-pointer -g -O1 \
  -Iinclude -Isrc -Ideps/clog/include -DCPUINFO_LOG_LEVEL=5 -DCPUINFO_ARCH_ARM64=1 \
  repro_cpulist.c src/linux/cpulist.c src/log.c deps/clog/src/clog.c -o repro_cpulist
```

Before (66ee79c):

```
==ERROR: AddressSanitizer: stack-buffer-overflow ... READ of size 1
    #0 parse_entry cpulist.c:119
    #1 cpuinfo_linux_parse_cpulist cpulist.c:198
  [32, 288) 'buffer' (line 157) <== Memory access at offset 288 overflows this variable
```

After the patch it's clean under ASan, logs `missing end of range in the cpu list entry "..."` and returns false.

There's no cpulist unit test in the tree today so I left this out of the PR, but I'm happy to wire the case into the mock setup if you want it covered.

soma0212:
any update here?

PR #435, opened 2026-09-21
Add Arm Neoverse V3 recognition

ashuie:
Support the Arm Neoverse V3 architecture.

Verified on a Neoverse V3 `gv5` host:
  - MIDR: `0x410fd841` (`r0p1`)
  - L1 data: 64 KiB, 4-way
  - L1 instruction: 64 KiB, 4-way
  - private L2: 2048 KiB, 8-way per core
  
  For reference: Adding Neoverse V2 support: https://github.com/pytorch/cpuinfo/pull/194


Rohanjames1997:
@malfet @zhili03 @fbarchard - could you take a look at this PR? Thanks!

src/arm/linux/aarch64-isa.c
```diff
@@ -65,6 +66,7 @@ void cpuinfo_arm64_linux_decode_isa_from_proc_cpuinfo(
 			case UINT32_C(0x4100D400): /* Neoverse V1 */
 			case UINT32_C(0x4100D490): /* Neoverse N2 */
 			case UINT32_C(0x4100D4F0): /* Neoverse V2 */
+			case UINT32_C(0x4100D840): /* Neoverse V3 */
```
nSircombe: This will pick up V3 but mis V3AE.
It should also include an entry for V3AE.
I believe the correct MIDR would be `0x410FD830` (Implementor = 0x41, part = 0xD83).

see https://github.com/OpenMathLib/OpenBLAS/pull/5759

src/arm/linux/aarch64-isa.c
```diff
@@ -110,6 +112,7 @@ void cpuinfo_arm64_linux_decode_isa_from_proc_cpuinfo(
 		case UINT32_C(0x4100D490): /* Neoverse N2 */
 		case UINT32_C(0x4100D4A0): /* Neoverse E1 */
 		case UINT32_C(0x4100D4F0): /* Neoverse V2 */
+		case UINT32_C(0x4100D840): /* Neoverse V3 */
```
nSircombe: This will pick up V3 but mis V3AE.
It should also include an entry for V3AE.
I believe the correct MIDR would be 0x410FD830 (Implementor = 0x41, part = 0xD83).

see OpenMathLib/OpenBLAS#5759

nSircombe (changes requested):
This will pick up V3 but miss V3AE. Could you add the MIDR for v3ae? (0x410FD830, I think - Implementor = 0x41, part = 0xD83).

see OpenMathLib/OpenBLAS#5759

ashuie:
@nSircombe Thanks, I've added V3AE.

